PR #1, opened 2015-04-28
Create partners

mconnell:
Create a Partner object and a basic UI for adding/editing.


paulanthonywilson:
I'm assuming it's ok not to internationalise the admin stuff, as it's internally facing.

:+1:

PR #3, opened 2015-04-28
Add contacts

mconnell:
This PR adds basic functionality to add / remove contacts to a Partner. We've not had any design work done for this project so views are purely functional.


app/views/partners/_form.html.erb
```diff
@@ -22,6 +22,21 @@
...
+  <%= form.fields_for :contacts do |contact_form| %>
+    <div class="webfeeds__form-row">
+      <%= contact_form.label :name %>
+      <%= contact_form.text_field :name %>
+    </div>
+    <div class="webfeeds__form-row">
+      <%= contact_form.label :email %>
+      <%= contact_form.text_field :email %>
+    </div>
+    <%= contact_form.link_to_remove "Remove this contact" %>
```
JiggyPete: Given the validation on the model that there has to be at least one contact, I would suggest hiding this link on the first (or last remaining) contact so it can't be removed.

spec/models/contact_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+RSpec.describe Contact, :type => :model do
+
+  describe "validations" do
+    it "must have a name" do
+      expect(subject).to be_invalid
+      expect(subject.errors[:name]).to include("can't be blank")
+    end
+
+    describe "email" do
```
JiggyPete: Do we need all of these examples given it is testing a gem which we trust?
Or are these examples for confidence that the gem does it's job?

Could it all be packed down to 1 or 2 examples to demonstrate the gem is being used?

app/views/partners/_form.html.erb
JiggyPete: At the same time is it worth it if the UI is in flux?

app/views/partners/index.html.erb
```diff
@@ -16,6 +16,7 @@
         <td><%= partner.name %></td>
         <td><%= partner.url %></td>
         <td><%= link_to 'Edit', [:edit, partner] %></td>
+        <td><%= link_to 'Show', partner %></td>
```
JiggyPete: Do we need to worry about webfeeds being multi-lingual?
I'd suggest getting this in as early as possible if it is. Going by the name it's probably internal only though?

seane: The man from Partnerships (Ed), he say 'yeah'! His pref is for EN and CY. Hope that's OK

mconnell: @seane Are there any internal users that would use a Welsh version? Our understanding is that the tool will be able to publish English and Welsh feeds but there are only going to be a small set of internal users using the admin functionality?

spec/models/contact_spec.rb
urfolomeus: Yeah probably, but I don't see there harm in ensuring that the gem does what we expect. Personally I'd leave it in. :)

app/views/partners/_form.html.erb
urfolomeus: UI is in flux, but probably a good idea to hide the Remove button if there's only one contact. That said, it would have to be done in JavaScript so that the button was removed if you started with multiple contacts and removed until you only had one left.

Either way, better to leave until UI is being worked on.

app/views/partners/index.html.erb
seane: Hey @mconnell Gotcha - internally for admin purposes, EN is sufficient

mconnell: ok doke, thanks!

paulanthonywilson:
:+1:

JiggyPete:
:+1:

PR #4, opened 2015-04-29
Create feeds

mconnell:
Adds the ability to add a basic Feed object to a partner, and the ability to edit feed objects.

Note: There is a removal of a validation and a spec that appears as a result of rebase shenanigans because we branched from a branch before everything had been merged in (due to PRs). The validation was intentionally removed


app/views/feeds/_form.html.erb
```diff
@@ -0,0 +1,30 @@
+<% if feed.errors.any? %>
+  <div id="webfeeds__error-explanation">
```
neoeno: Is this a global BEM prefix for all classes in the project?

app/views/feeds/_form.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <div class="webfeeds__form-row">
+    <%= form.label :name %>
+    <%= form.text_field :name %>
+  </div>
+  <div class="webfeeds__form-row">
+    <%= form.label :description %>
+    <%= form.text_field :description %>
+  </div>
+
+    <div class="webfeeds__form-row">
```
neoeno: Indentation is off.

app/views/feeds/_form.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <%= link_to 'cancel', partner %>
+    <%= form.submit %>
+  </div>
+<% end -%>
```
neoeno: Who cares, but recent Rails enable whitespace suppression by default, making this a no-op. [See here.](https://github.com/rails/rails/blob/fcbdac7e82725c388bf5adf56a9a9a16d4efdbe0/actionview/lib/action_view/template/handlers/erb.rb#L77)

app/views/partners/show.html.erb
```diff
@@ -4,8 +4,23 @@
...
+
+<%= link_to 'Add Feed', new_partner_feed_path(@partner) %>
+
+<% @partner.feeds.each do |feed| %>
+  <div>
+    <%= feed.name %>
+    <br>
+    <%= feed.description %>
+    <%= link_to 'Edit feed', edit_partner_feed_path(@partner, feed) %>
+  </div>
```
neoeno: I'm guessing this is temporary scaffold-y stuff?

mconnell: Yeah we're expecting it all to change once we have some design/UX love

app/views/feeds/_form.html.erb
mconnell: it's a BEM prefix of sorts. I started with webfeeds__ just to get us going, if there is something better I'm open to suggestions of how we should do it?

neoeno: It's more that if this is going to be a prefix on every class, it could turn into an unnecessary hassle — given that this project is standalone anyway. Your call though!

benbarnett: I'd agree actually, if this is standalone then the namespace might be overkill

mconnell: I'll take it out then!

neoeno: Also it's an ID :trollface:

neoeno:
:+1: feed the web

darokel:
Just one comment, but apart from that :+1:

PR #5, opened 2015-04-30
Ingest blog feed

mconnell:
First pass at reading and capturing data from the MAS blog feed.

* Creates Source, SourceItem, and SourceDocument models.
* Setups a seed for the MAS blog RSS.
* Define a method Source#fetch! to import/update data from the feed.


app/models/source.rb
```diff
@@ -0,0 +1,27 @@
...
+  def update_or_create_source_item!(item_attributes)
+    source_item = SourceItem.find_or_initialize_by(
+      source_id: self.id,
+      guid: item_attributes[:guid]
+    )
+
+    source_item.title        = item_attributes[:title]
+    source_item.body         = item_attributes[:body]
+    source_item.url          = item_attributes[:url]
+    source_item.published_at = item_attributes[:published_at]
```
neoeno: `[:title, :body, :url, :published_at].map{ |field| source_item.send field, item_attributes.try(:fetch, l, nil) }.compact`

much better

urfolomeus: http://media.giphy.com/media/j3iGKfXRKlLqw/giphy.gif

spec/rails_helper.rb
```diff
@@ -48,3 +49,7 @@
   # https://relishapp.com/rspec/rspec-rails/docs
   config.infer_spec_type_from_file_location!
 end
+
+def fixture(filename)
+  IO.read(Rails.root.join("spec", "fixtures", filename))
+end
```
neoeno: why have I never done this

neoeno:
:+1:

darokel:
:+1: I left a couple of comments but found you already addressed what I raised in later commits. So I deleted my notes :)

PR #6, opened 2015-05-04
Allow admins to select content from the blog to use in a feed

mconnell:
Adds the ability to add/update content used in a feed. Also introduces factory girl to make things a bit more readable.


app/controllers/feeds_controller.rb
```diff
@@ -1,5 +1,9 @@
 class FeedsController < ApplicationController
   before_action :find_partner
+  before_action :find_feed, only: [:show, :edit, :update]
```
paulanthonywilson: ```except: :new``` ?

spec/models/contact_spec.rb
```diff
@@ -1,13 +1,20 @@
 require 'rails_helper'
 
 RSpec.describe Contact, type: :model do
+  it 'has a valid factory' do
```
paulanthonywilson: Is this a test configuration test?

app/models/feed.rb
```diff
@@ -1,5 +1,21 @@
...
+  def update_source_items(source_item_ids)
+    FeedSourceItem.transaction do
+      feed_source_items.destroy_all
+
+      begin
+        source_item_ids.each do |source_item_id|
+          feed_source_items.create!(source_item_id: source_item_id)
+        end
+      rescue
+        raise ActiveRecord::Rollback
```
neoeno: I could be wrong, but I think any exception will trigger the rollback. So this begin & rescue might be unnecessary — since the `create!` will raise an exception anyway.

Try this and watch the SQL logging (frontend's `rails c` has it on)

```
FeedSourceItem.transaction do
  FeedSourceItem.find(4643265657354) # no record probably
end
```

That is, unless you're anticipating another sort of exception from this block?

spec/models/feed_spec.rb
```diff
@@ -1,20 +1,136 @@
...
+          expect(subject.source_items.count).to eql(1)
+          expect(subject.source_items).to include(source_item)
+        end
+      end
+
+      context 'when given bad data' do
+        it 'rollsback the entire transaction' do
+          source_item_ids = [source_item.id, nil]
+          subject.update_source_items(source_item_ids)
+          expect(subject.source_items).to be_empty
```
paulanthonywilson: I see that it's demonstrating that the first source is not added if the 2nd fails.  I wonder if it's worth demonstrating that any existing source item ids that are destroyed, are retained.

spec/models/partner_spec.rb
```diff
@@ -1,13 +1,19 @@
 require 'rails_helper'
 
 RSpec.describe Partner, type: :model do
+  it 'has a valid factory' do
+    expect(FactoryGirl.build(:partner)).to be_valid
```
paulanthonywilson: Spec configuration specs are strange and interesting.

app/models/source_item.rb
```diff
@@ -1,8 +1,12 @@
 class SourceItem < ActiveRecord::Base
   belongs_to :source
+  has_many :feed_source_items, dependent: :destroy
+  has_many :feeds, through: :feed_source_items
 
   validates :title, :body, :url, :published_at, :guid, presence: true
 
+  scope :by_published_at, -> { order('published_at DESC') }
```
neoeno: Minor point, but I prefer `order(created_at: :desc)` — partially because it behaves more predictably in table joins (which might be more likely on a scope).

app/views/feed_source_items/edit.html.erb
```diff
@@ -0,0 +1,15 @@
...
+      <%= check_box_tag "source_item_ids[]", source_item.id, @feed.source_items.include?(source_item) %>
+      <%= label_tag source_item.title %>
+    </div>
+  <% end %>
+
+  <div class="webforms__form-actions">
+    <%= link_to 'cancel', [@partner, @feed] %>
+    <%= submit_tag 'Save' %>
+  </div>
+<% end %>
```
neoeno: Is there a reason these are written in whatever this style is called? Rather than whatever the `form_for` style is called?

app/views/feeds/_form.html.erb
```diff
@@ -24,7 +24,7 @@
   </div>
 
   <div class="form-row">
-    <%= link_to 'cancel', partner %>
+    <%= link_to 'cancel', feed.persisted? ? [partner, feed] : partner %>
```
neoeno: Are these `[partner, feed]` things synonymous with `path_for` helpers? I find the helpers easier to read — what do you think?

app/views/feeds/show.html.erb
```diff
@@ -0,0 +1,14 @@
+<h1><%= @feed.name %></h1>
+<p><%= @feed.description %></p>
+
+<%= link_to 'Edit feed', edit_partner_feed_path(@partner, @feed) %>
+
+<h2>Content</h2>
+
+<%= link_to 'Add content', [:edit, @partner, @feed, :feed_source_items] %>
```
neoeno: Likewise here.

app/views/feeds/show.html.erb
```diff
@@ -0,0 +1,14 @@
...
+
+<ul>
+  <% @feed.source_items.by_published_at.each do |item| %>
```
neoeno: I like this use of scopes.

app/views/partners/show.html.erb
```diff
@@ -21,6 +21,6 @@ Url: <%= link_to @partner.url, @partner.url %>
     <%= feed.name %>
     <br>
     <%= feed.description %>
-    <%= link_to 'Edit feed', edit_partner_feed_path(@partner, feed) %>
+    <%= link_to 'Show feed', [@partner, feed] %>
```
neoeno: Similarly here — the array style has been actively preferred — I'm guessing it's a deliberate decision?

app/controllers/feeds_controller.rb
mconnell: There's a `:create` action as well. So we could blacklist `[:new, :create]` and that would work. But if   someone comes along later and adds another action, they'll automagically get the behaviour only intended for the `:show`, `:edit`, and `:update` actions. So for that reason i'm inclined not to change it

app/views/feeds/show.html.erb
mconnell: Yes, they are shorthand helpers that have existing in rails since the dawn of time. I have a preference for these when dealing with nested resources as the helper this generates is `edit_partner_feed_feed_source_items_path(@partner, @feed)`. I think it's a bit long and harder to read than the short hand version. The one above was done at the same time to be consistent.

app/views/partners/show.html.erb
mconnell: yerp, see comment above.

app/views/feeds/show.html.erb
mconnell: :fireworks:

app/models/source_item.rb
mconnell: Yup, forgot you can now use a hash in `order` now

app/controllers/feeds_controller.rb
paulanthonywilson: Fair enough.

(TBH, I’m not overly keen on filters for this stuff. I’d always preferred the Decent Exposure approach, but I remember that everyone else hates it.)

app/views/feed_source_items/edit.html.erb
mconnell: I think this was the first way that came to mind, a `form_for` would require us to specify the controller/action in addition to `[@partner, @feed]` and maybe some jiggery-pokery with `fields_for` to get the through association on SourceItem objects to play nicely.

We can look at changing it if there is a strong preference?

neoeno:
:+1:

app/views/feed_source_items/edit.html.erb
neoeno: Just discovered this, reckon I'll use it somewhere someday http://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html#method-i-collection_check_boxes

neoeno: Oh and, for anyone reading along at home — we discussed it in office and I've no strong preference.

app/views/feed_source_items/edit.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1>Edit Content</h1>
+
+<%= form_for [@partner, @feed], url: [@partner, @feed, :feed_source_items] do |form| %>
+
+  <%= form.collection_check_boxes :source_item_ids, @source_items, :id, :title do |b| %>
+    <div class="checkbox">
+      <%= b.label { b.check_box + b.text } %>
```
neoeno: Sorry, more comments!

This method of labelling checkboxes (putting them in the label) is deprecated — I only know because it came up in the RAD DAC report recently. Using the explicit `for=` is recommended. So something like:

```
<%= b.check_box %>
<%= b.label b.text %>
```

If Rails sets up the labels' `for=` correctly?

mconnell: on it

munckymagik:
:+1:

PR #7, opened 2015-05-05
Move admin behaviour to admin namespace.

mconnell:
Not a terribly exciting PR. We've decided that the admin functionality should be moved into an admin namespace to enforce a separation of concerns.


darokel:
:+1:

Tr4pSt3R:
:+1: good idea

mconnell:
Thanks! :star2:

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,5 @@
+class AdminController < ApplicationController
+  if Rails.env.production?
+    http_basic_authenticate_with name: ENV['BASIC_AUTH_NAME'], password: ENV['BASIC_AUTH_PASSWORD']
```
tomas-stefano: I know that the repo already have the ENV explanation [here](https://github.com/moneyadviceservice/webfeeds#deployment) but I can't see the dot-env or env relate gems in the Gemfile. If I want to use this env variables, do I need to export by hand or do some command? If so, can you add the command setup for that? (or even if is just a copy command). e.g add `cp .env-dev .env` to readme

PR #9, opened 2015-05-11
Bourbon

spencer:
Add Bourbon, Font-Awesome and basic first pass styling.


app/assets/javascripts/application.js
```diff
@@ -14,3 +14,17 @@
 //= require jquery_ujs
 //= require jquery_nested_form
 //= require_tree .
+
+$(document).ready(function() {
```
urfolomeus: Should we put this javascript in a separate file and require it in here?

app/assets/stylesheets/application.css.scss
```diff
@@ -0,0 +1,175 @@
...
+        color: $navigation-color-hover;
+      }
+    }
+  }
+
+  .active-nav-item a {
+    border-bottom: 1px solid $navigation-active-link-color;
+    padding-bottom: 3px;
+  }
+}
```
urfolomeus: Same comment here. I'm not sure what the house style is but I generally prefer to keep application.(css|js) for requires only. Placing styles into a separate style file and then requiring in here.

benbarnett:
Is there a reason we're not using any of the MAS styleguide for this?

app/assets/stylesheets/application.css.scss
spencer: Sure - It's WIP to show some visual progress...as we were just starting.

app/assets/javascripts/application.js
spencer: Again, WIP, so yes, just hacking stuff in for the purpose of a start.

spencer:
@benbarnett As it's not a client facing project (only ever used internally to generate content which is consumed), styling it to look different from the MAS website seemed OK? Wrong?

benbarnett:
@spencer We use Dough and the like on our CMS which is also internally facing. The benefits aren't just for external users – having consistent coding standards is incredibly useful for ongoing work too. On Contento (CMS), we're evening gaining a couple of Dough components that we can carry across to the public site just because things were built using the same standards.

Understand this is WIP but would be great if the front-end guidelines are followed for both public and non-public facing work.

darokel:
As per @benbarnett's comments :+1:

PR #10, opened 2015-05-12
Add pixel tracking

urfolomeus:
We're waiting on direction for which API we need to use to attach to Google Analytics (to be decided on at a meeting tomorrow). Until then we've put the basic structure in place for the pixel tracker generation, insertion and serving.


spencer:
Looks good to me :+1:

app/views/admin/feeds/show.html.erb
```diff
@@ -1,14 +1,14 @@
...
 
-<h2>Share</h2>
-<div>
-  <h3>RSS Feed URL</h3>
-  <%= text_field_tag :rss, feed_url(@feed, format: :rss) %>
-  <%= link_to 'Open in a new tab', feed_url(@feed, format: :rss), target: '_blank' %>
-</div>
+<h2>Rules</h2>
+<ul>
+  <li>Pixel tracking - <%= @feed.pixel_tracked? ? '✓' : '✗' %></li>
```
neoeno: Nice bit of UTF-8 here!

app/views/admin/feeds/show.html.erb
```diff
@@ -19,3 +19,11 @@
...
   <% end %>
 </ul>
+
+
+<h2>Share</h2>
+<div>
+  <h3>RSS Feed URL</h3>
+  <%= text_field_tag :rss, feed_url(@feed, format: :rss) %>
+  <%= link_to 'Open in a new tab', feed_url(@feed, format: :rss), target: '_blank' %>
+</div>
```
neoeno: This div element not strictly necessary here — though if this is still just temporary markup it's not worth worrying about I suppose.

app/views/admin/feeds/show.html.erb
urfolomeus: Thanks. It's temporary until the design's in place :)

app/models/pixel_tracker.rb
```diff
@@ -0,0 +1,35 @@
+class PixelTracker
```
munckymagik: If this isn't strictly a model, is there another folder it could/should go in? Are there likely to be more of these type of classes that can be collected together, say in `lib` maybe? If this is a one-off then it seems OK to me to leave it where it is.

app/views/admin/feeds/show.html.erb
urfolomeus: Correct and correct :)

docs/publish2_pixel.txt
```diff
@@ -0,0 +1,55 @@
...
+  partner_id=4745&
+
+http://www.publish2.com/static/tracking?
+  partner_name=Right_Move&
+  story_id=9626775&
+  story_title=06052015+-+RightMove+-+Pension+Reforms&
+  partner_feed_id=3381&
+  distributor_feed_id=2259&
+  path=%2Fen%2Fbespoke%2Frightmove%2Fpension-reforms%3Fpublish2_story_id%3D9626775%26publish2_byline%3DNone%26publish2_categories%3Drightmovepensions%2Bcustom_feed_3381%26publish2_partner_name%3DRight_Move%26publish2_partner_id%3D4923&
+  partner_id=4923&
```
neoeno: Did you mean to check this in?

app/models/pixel_tracker.rb
urfolomeus: We now have a folder called `rules` that this and others reside in (although it too is being refactored). Not an issue IOW :)

JiggyPete:
:+1: Looks good, well tested. 
Got it all running.
Really like pixel_tracker.rb

docs/publish2_pixel.txt
urfolomeus: Yes. It may get pulled out again later, but we need it until we know the direction the pixel tracking is going to take.

spec/models/pixel_tracker_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'rails_helper'
+
+describe PixelTracker do
+  let(:regexp) { Regexp.new('<img src="(.*)" \/>') }
```
neoeno: Munging HTML with regex is one of my pet hates. That said, if you truly are expecting no other attributes and exactly this form...

urfolomeus: Yeah we held off on this until it was causing too much pain _not_ to. Sorry :(

munckymagik:
:+1:

PR #12, opened 2015-05-15
Add custom rules

mconnell:
This PR allows an admin to apply custom rules to a feed:

* Prepend – prepend specific content to all items in a feed
* Append – append specific content to all items in a feed
* transform – apply a regular expression transform to all items in a feed.

We've also introduced karma/chai/mocha/phantomjs for JS testing (mimicking the RAD setup). And followed the dough pattern for creating a dough component for toggling a text field being enabled or not.


asmega:
the ruby stuff looks fine :+1:

app/assets/javascripts/modules/RuleForm.js
```diff
@@ -0,0 +1,71 @@
...
+   * Set up listeners
+   * @param  {[type]} initialised [description]
+   * @return {[type]}             [description]
+   */
+  RuleFormProto.init = function(initialised) {
+    var self = this;
+
+    this.$el.find('#rule_rule_type').on('change', function () {
+      self.toggleRegex();
+    });
```
neoeno: You could avoid the self, by doing:

```
this.$el.find('#rule_rule_type').on('change', $.proxy(this.toggleRegex, this));
```

`$.proxy` is basically `Function.prototype.bind`.

app/assets/javascripts/modules/RuleForm.js
```diff
@@ -0,0 +1,71 @@
...
+  RuleFormProto.toggleRegex = function() {
+    if(this.$ruleType.val() === 'transform'){
+      this.$ruleRegex.prop('disabled', false);
+    } else {
+      this.$ruleRegex.prop('disabled', true);
+    };
+  };
+
+  return RuleForm;
+});
```
neoeno: newline

app/assets/javascripts/require_config.js.erb
```diff
@@ -0,0 +1,21 @@
+//= depend_on_asset modules/RuleForm
+
+<%
+  def requirejs_path(asset)
+    javascript_path(asset).sub(/.js\z/, '')
+  end
+  requirejs_config = {
+     # baseUrl: (Rails.application.config.action_controller.asset_host || '') + \ Rails.application.config.assets.prefix,
```
neoeno: This intentional?

app/assets/stylesheets/_enhanced.scss
```diff
@@ -22,3 +22,18 @@
...
 
 body {background: #DDD;}
+
+.form__row label {
+  display: inline-block;
+  vertical-align: top;
+  min-width: 100px;
+}
+
+.form__row input[type=text] {
```
neoeno: Dough standard is `.form__control` or something, [files](https://github.com/moneyadviceservice/dough/blob/master/assets/stylesheets/components/common/_form.scss)

app/assets/javascripts/require_config.js.erb
mconnell: The comment? We pretty much copy pasted the config from RAD to get the dough JS stuff working so I hadn't noticed it.

app/views/admin/rules/_form.html.erb
```diff
@@ -0,0 +1,42 @@
+<% if rule.errors.any? %>
+  <div id="error-explanation">
```
neoeno: underscores for IDs I think

spec/javascripts/karma.conf.js
```diff
@@ -0,0 +1,40 @@
+module.exports = function(config) {
```
neoeno: I stopped reviewing here because I want to go home. I'll continue.

app/assets/javascripts/modules/RuleForm.js
urfolomeus: What's the danger in using `self`? Seems more explicit to me.

neoeno: No danger as such — it's just not necessary. The `$.proxy`/`.bind` style is more 'modern', also I suppose. 

If you do prefer it, generally if I define a self I use it throughout the scope it's in, rather than alternating between them — it can cause confusion.

spec/javascripts/karma.conf.js
```diff
@@ -0,0 +1,40 @@
...
+    exclude: [
+      'vendor/assets/bower_components/**/test/**/*.js',
+      'vendor/assets/bower_components/**/spec/**/*.js'
+    ],
+
+    // preprocess matching files before serving them to the browser
+    // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
+    preprocessors: {
+      '**/*.html': ['html2js']
+    },
```
neoeno: What's this for? I've only seen it used in AngularJS projects. Might well have a use here, though.

app/assets/javascripts/modules/RuleForm.js
urfolomeus: Good point on the self/this flipping. Not sure I agree on using proxy just because it's more modern. :)

README.md
```diff
@@ -21,6 +21,11 @@ fetch dependencies:
 
 To run the tests, simply invoke `rake`. This will run the test suite, as well as rubocop. The rubocop configuration is a copy of the configuration used in [frontend](https://github.com/moneyadviceservice/frontend/blob/master/.rubocop.yml) to match the MAS style.
 
+We also have some javascript tests which can be run separately using:
+```bash
+  node_modules/.bin/karma start spec/javascripts/karma.conf.js --single-run=true
```
benbarnett: Can be cleaner to add these to `package.json` and then running `npm test`. Like this: https://github.com/moneyadviceservice/dough/blob/master/package.json#L7

app/assets/javascripts/application.js
```diff
@@ -1,16 +1,13 @@
...
-// about supported directives.
-//
 //= require jquery
 //= require jquery_ujs
 //= require jquery_nested_form
-//= require_tree .
+//= require require_config.js.erb
+
+// Components
+require(['jquery', 'componentLoader'], function ($, componentLoader) {
```
benbarnett: Remove space after `function` and before `(`

app/assets/javascripts/application.js
```diff
@@ -1,16 +1,13 @@
...
+  componentLoader.init($('body'));
+});
+
+require(['jquery'], function ($) {
```
benbarnett: Same here re space

app/assets/javascripts/modules/RuleForm.js
neoeno: It's your call — different MAS projects use different styles I think. In other languages, we might think having two references to the same object that we use in different places could be confusing. But this isn't other languages!

neoeno:
fyi — https://www.moneyadviceservice.org.uk/en/styleguide/javascript — the `.jshintrc` will help with the style stuff.

app/assets/javascripts/application.js
urfolomeus: `function` is not a function though, it's a keyword.

app/assets/javascripts/application.js
urfolomeus: We put space after anonymous calls to `function` for the same reason that we put them in `if (condition)` and `for (loop_stuff)`.

app/assets/javascripts/modules/RuleForm.js
```diff
@@ -0,0 +1,71 @@
...
+
+  'use strict';
+
+  var RuleFormProto,
+      defaultConfig = {};
+
+  function RuleForm($el, config) {
+    RuleForm.baseConstructor.call(this, $el, config, defaultConfig);
+
+    this.$ruleRegex = this.$el.find('#rule_regex').first();
```
benbarnett: Try to avoid using ids as selectors. You can use a css class with a `js-` prefix

app/assets/javascripts/application.js
benbarnett: I can see the reasoning but just for consistency we follow this: https://github.com/moneyadviceservice/javascript#functions

neoeno:
Oh, yes, and before I forget — javascript tests :ok_hand:

app/assets/javascripts/application.js
urfolomeus: Fair enough :)

app/views/admin/rules/_form.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+    <ul>
+      <% rule.errors.full_messages.each do |msg| %>
+        <li><%= msg %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
+
+<%= form_for [:admin, partner, feed, rule], data: {dough_component: "RuleForm"} do |form| %>
```
benbarnett: We have a Dough helper for forms. Example here: https://github.com/moneyadviceservice/rad/blob/master/app/views/principals/new.html.erb#L6-L7

app/views/admin/rules/_form.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+  <div class="form__row">
```
benbarnett: If you use the Dough helper these can be generated for you, a la https://github.com/moneyadviceservice/rad/blob/master/app/views/principals/new.html.erb#L6-L7

app/views/admin/rules/_form.html.erb
urfolomeus: Thanks @benbarnett.

app/views/admin/rules/_form.html.erb
urfolomeus: Grand we'll take a look at that. Going forward, how do we know this stuff is there? (apologies if this is just me being noob) :)

urfolomeus: Ignore that @benbarnett. Found the styleguide :)

app/views/admin/rules/_form.html.erb
benbarnett: Forgot to mention, the Dough form things also takes care of this validation summary: https://github.com/moneyadviceservice/rad/blob/master/app/views/principals/new.html.erb#L8

We can definitely improve the documentation for this kind of thing

JiggyPete:
:+1: all looks good. Was a bit lost as to all the comments in the RuleForm.js?

spec/javascripts/karma.conf.js
mconnell: Is there something specific in this file you're referring to? The karma runner won't run without the above content commented out.

app/assets/javascripts/modules/RuleForm.js
mconnell: "The id attribute and js-* class names are reserved for JavaScript-only use." [CSS Styleguide](https://www.moneyadviceservice.org.uk/en/styleguide/css)

If there is a preference for `js-` prefixed classes, should the style guide be updated?

benbarnett: Yes, that's badly worded. It states `js-` at the top and then contradicts itself further down. I'll fix.

PR #13, opened 2015-05-19
Ingest contento articles

mconnell:
This PR allows us to fetch article content from contento in addition to content from the blog.

* Renamed `SourceDocument` to a more explicit name `XmlSourceDocument`
* Introduced a `JsonSourceDocument`
* Added a `content_type` attribute to `Source`
* Moved some fixtures into folders to tidy things up.


asmega:
:+1:

app/models/json_source_document.rb
```diff
@@ -0,0 +1,19 @@
+require 'open-uri'
+
+class JsonSourceDocument
+  def initialize(url)
+    @document = JSON.parse(open(url).read)
```
neoeno: Just to check — is `url` (& what it returns) trusted? Both to be non-malicious, and valid.

app/models/json_source_document.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def items
+    @document.map do |json_item|
+      {
+        title:        json_item['label'],
+        body:         json_item['blocks'][0]['content'],
+        url:          "https://www.moneyadviceservice.org.uk#{json_item['full_path']}",
```
neoeno: Should this URL be a constant? Or configurable?

app/models/json_source_document.rb
mconnell: The url comes from an instance of Source, which is currently a class non-accessible to the admin user, and are defined in seeds.rb. All source urls are strings representing data sources for mas content eg. The blog, contento

app/models/json_source_document.rb
mconnell: It could be a constant. It doesn't need to be configurable.

urfolomeus:
I like the command pattern used to select the right parser based on the content type. I also think it's a sensible choice to use webmock to simulate the actual parsing in the tests. It makes it easy to see the process.

:+1:

JiggyPete:
:+1:

PR #14, opened 2015-05-19
breadcrumb navigation

mconnell:
This PR just adds some simple breadcrumbs (no rails helpers yet) to make navigating around a little bit less painful and to steer things towards the mockups spencer did.


app/assets/stylesheets/_layout.scss
```diff
@@ -12,6 +12,10 @@
   header {
     @include column(12);
     padding: $baseline-unit*4 0;
+
+    a {
+      color: white;
+    }
```
neoeno: This selector (everything within `.l-header`) is starting to look like it needs a restructure. Does it make sense to have a semantic `header` element within an `.l-header`? What does `a` mean here? I'll probably comment on the DOM section when I get down that far.

app/views/admin/partners/show.html.erb
```diff
@@ -1,4 +1,7 @@
-<%= link_to 'Back to partners', admin_partners_path %>
+<ul class="breadcrumb">
+  <li><%= link_to 'Partners', admin_partners_path %></li>
+  <li><%= @partner.name %></li>
+</ul>
```
neoeno: This breadcrumbs thing seems like a good candidate for a helper or a partial helper.

app/views/layouts/application.html.erb
```diff
@@ -19,7 +19,7 @@
 <body>
 <div class="l-header">
   <header>
-    Webfeeds
+    <%= link_to 'Webfeeds', root_url %>
   </header>
 </div>
```
neoeno: See comment at the top. e.g.

```
<header class="l-header">
  <%= link_to 'Webfeeds', root_url %>
</header>
```

If you're going to just have the one link.

munckymagik:
What Caden said, otherwise :+1:

neoeno:
What Dan said :+1:

JiggyPete:
:+1:

app/views/admin/partners/show.html.erb
benbarnett: Here here. If you do go this route, then could you add a class to the `li` and style them rather than the element directly? Its a BEM thing – perhaps `breadcrumb__item` would do it. If not don't worry too much, I know this isn't customer facing.

PR #15, opened 2015-05-20
rake task webfeeds:fetch_all

mconnell:
This adds a rake task `rake webfeeds:fetch_all`. This task is intended to be used by cron or other scheduling tool to periodically poll the different sources for new content.


neoeno:
Agree with @JiggyPete on the `fetch_all` thing — seems like it's gaining responsibility for logging etc. I've no objection to there being a fetch_all method on Source though, just that maybe everything other than 'fetch all the sources' should be done elsewhere.

Otherwise, looks great :+1:

JiggyPete:
:+1: awesome!

PR #16, opened 2015-05-22
Firehose feed

mconnell:
This PR introduces the notion of 'Firehose' feeds. Existing feeds are now 'Curated'.

A curated feed is one where the admin has selected specific content to be pushed out as a new feed.
A firehose feed outputs all of the content from a given source. Any new content from a source should automatically be outputted by a firehose feed.

Both types of feeds may have rules applied to them.


app/controllers/admin/firehose_feed_sources_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    else
+      flash[:alert] = 'Failed to update sources.'
+    end
+    redirect_to admin_partner_feed_path(@partner, @feed)
+  end
+
+  private
+
+  def allowed_params
+    params[:feed][:source_ids].pop
```
neoeno: What's this do? Bump off some rails checkbox thing?

mconnell: yarp

neoeno:
:+1:

app/controllers/admin/firehose_feed_sources_controller.rb
munckymagik: Could leave a comment to explain it maybe? E.g. https://github.com/moneyadviceservice/rad/blob/master/app/forms/admin/move_advisers_form.rb#L13

mconnell: Yep, can do.

app/models/feed.rb
```diff
@@ -1,22 +1,9 @@
...
-
-        source_item_ids.each do |source_item_id|
-          feed_source_items.create!(source_item_id: source_item_id)
-        end
-      rescue
-        raise ActiveRecord::Rollback
-      end
-    end
-  end
+  validates :type, inclusion: { in: VALID_TYPES, message: "must be one of #{VALID_TYPES.join(', ')}" }
```
munckymagik: Are we using input validation to check an issue of code-correctness? If so, could this be guaranteed by testing/specing for code correctness instead?

munckymagik:
:+1:

app/models/feed.rb
urfolomeus: We're doing this to prevent bad data from getting into the database, regardless of whether or not we should be doing more testing at other points where this data is being used. :) (IMHO)

PR #19, opened 2015-05-22
Pixel tracking - Google analytics

mconnell:
This PR adds the google analytics portion of work to the pixel tracker that may be appended to each item in an outputted feed.

When our pixel tracking receives a request. It submits a pageview to google analytics via the [GA Measurement Protocol](https://developers.google.com/analytics/devguides/collection/protocol/v1/). See the `Analytics` class for details.

Additionally, when we submit information to google analytics, we append a number of additional parameters to the url. These parameters are used by the internal Customer Insight Database tool, and are required to mimic the values currently send by the existing system. An extra parameter `webfeeds` with the value `'1'` is also supplied so we have a way of indicating the analytics were provided by this source.

These parameters could potentially be supplied as custom metrics/dimensions within google analytics in the future, but is out of scope for the moment.


app/controllers/tracking_controller.rb
```diff
@@ -2,10 +2,31 @@ class TrackingController < ApplicationController
...
+
+  def analytics_options
+    options = params.dup
+    options.delete(:action)
+    options.delete(:controller)
+    options.delete(:format)
+
+    options.merge(
+      ip_address: request.remote_ip,
+      language:   request.accept_language[/[\w]{2}-[\w]{2}/i],
```
neoeno: Would we ever expect to get 2 letter codes here? Like `en`?

app/models/analytics.rb
```diff
@@ -0,0 +1,35 @@
+class Analytics
+  ENDPOINT = 'http://www.google-analytics.com/collect'
```
neoeno: Does this support HTTPS?

app/models/analytics.rb
```diff
@@ -0,0 +1,35 @@
...
+    form_params = {
+      'cid' => SecureRandom.uuid,
+      'dl'  => url_with_additional_params(options),
+      'dt'  => options[:source_item_title],
+      't'   => 'pageview',
+      'tid' => ENV['GA_TRACKING_ID'],
+      'ua'  => options[:user_agent],
+      'uip' => options[:ip_address],
+      'ul'  => options[:language],
+      'v'   => '1'
```
neoeno: I do love google's approach to naming.

app/models/analytics.rb
```diff
@@ -0,0 +1,35 @@
...
+  def self.url_with_additional_params(options)
+    params = {
+      'publish2_story_id'     => options[:source_item_id],
+      'publish2_byline'       => 'None',
+      'publish2_partner_name' => options[:partner_name],
+      'publish2_partner_id'   => options[:partner_id],
+      'webfeeds'              => '1'
+    }
+
+    params_string = params.map { |k, v| "#{k}=#{URI.escape(v)}" }.join('&')
```
neoeno: http://apidock.com/rails/Hash/to_query Might be useful

app/controllers/tracking_controller.rb
mconnell: The existing analytics only show 'en-us', 'fr-fr' etc.. And the API if I recall is expecting a string of 5 bytes.

mconnell: characters/bytes/things

munckymagik:
:+1:

neoeno:
:+1:

app/models/analytics.rb
mconnell: There is an HTTPS endpoint if we chose to use it. The request is being sent from the server to analytics, and all the information that we're sending is readily available in the params of the tracking .gif so I'm not sure if we gain anything by switching to it?

app/models/analytics.rb
mconnell: Haven't seen that before. thanks.

app/models/analytics.rb
neoeno: I tend to use it by default these days — if it's just an 's' different. If it's extra set-up though I wouldn't bother.

PR #22, opened 2015-05-25
Add categories to feeds

mconnell:
This PR introduces the notion of categories to the application.


asmega:
:+1:

app/presenters/source_item_presenter.rb
```diff
@@ -14,6 +14,9 @@ def to_xml
       xml.description { xml.cdata! body }
       xml.published_at @source_item.published_at.to_s(:rfc822)
       xml.guid @source_item.guid, isPermaLink: false
+      @source_item.categories.each do |category|
+        xml.category category.name
```
JiggyPete: I wouldn't have expected this to work for multiple categories, as it looks like xml.category will have the value of the last category.name, overwriting the other categories that have went before it.

The tests tell me I'm wrong and that it works :-)

JiggyPete:
:+1:

app/presenters/source_item_presenter.rb
urfolomeus: The DSL is a little confusing (especially if you're used to FactoryGirl's DSL for example) but `xml.category category.name` calls a method called `xml.category` which adds a  new `<category>` element each time. [edited by Mark ;)]

PR #25, opened 2015-05-28
Global rules concept

mconnell:
This PR provides the groundwork for us to add the ability for an admin to have global rules which apply to all feeds, and local rules that are per feed.

We're still discussing how the UI/UX of this will behave so created this PR to get the modelling aspect of it integrated.

In a nutshell, we've changed a relationship from a belongs_to/has_many, to a has_many via a join model. Which will allow us to reuse rules between different feeds. We've also added a flag to allow us to set a rule as global.


darokel:
:+1:

urfolomeus:
We did consider using Shoulda but weren't sure if you're using them on any other projects?

JiggyPete:
:+1:

paulanthonywilson:
:+1:

urfolomeus:
@paulanthonywilson I've heard it argued both ways but don't have a strong opinion myself. I do like to ensure that we have `dependent: :destroy` where we should though.

paulanthonywilson:
That thing when no-one reviews a PR for a while, so you think you'll pitch in. Then everyone does it at once.

paulanthonywilson:
@urfolomeus Yeah, but the difference between me and the people who argue it the other way, is that they are wrong. ;-)

darokel:
@urfolomeus Yeah we use it :)

PR #26, opened 2015-05-29
Manage sources

mconnell:
Adds the ability to add/edit sources


app/views/admin/sources/index.html.erb
```diff
@@ -0,0 +1,24 @@
+<h1>Sources</h1>
+
+<%= link_to 'Add Source', new_admin_source_path %>
+
+<table>
```
benbarnett: I'm sure this is fine as is but we have a table style you might want to consider: https://www.moneyadviceservice.org.uk/en/styleguide/components/common#section-scrollable-table

urfolomeus: Thanks, Ben. Looking into it now.

benbarnett:
if table seems fine then :+1:

jonalport:
:+1:

PR #27, opened 2015-05-29
move breadcrumb html into partial

mconnell:
Extracting breadcrumbs to a partial now that we have more uses of it.


jonalport:
:+1:

benbarnett:
Nice :+1:

spencer:
:+1:

PR #31, opened 2015-06-08
Validate rss

urfolomeus:
We had an issue where one of the fields that we'd added to the RSS feed was not valid RSS. To prevent this from happening again I've added RSS validation to the request specs for each feed.

This PR also makes the RSS being generated valid again (changing `published_at` for `pubDate`).


app/presenters/source_item_presenter.rb
```diff
@@ -12,7 +12,7 @@ def to_xml
       xml.title { xml.cdata! @source_item.title }
       xml.link { xml.cdata! @source_item.url }
       xml.description { xml.cdata! body }
-      xml.published_at @source_item.published_at.to_s(:rfc822)
+      xml.pubDate @source_item.published_at.to_s(:rfc822)
```
urfolomeus: This is the fix that was required :point_up:

spencer:
:+1:

asmega:
:+1:

PR #32, opened 2015-06-09
Edit rules ui

urfolomeus:
Apologies in advance for the largeness of this PR.

After a number of rounds of discussions around the rules that can be applied to feeds, we decided that we needed to refactor what we currently had in order to continue. Rather than do this in steps we decided to do this in one PR as we couldn't find smaller steps that made sense that weren't breaking. There are still more steps required in order for it to fully make sense, especially from a naming point of view.

The meat of this PR is to refactor out the concept of local rules (i.e. rules that only apply to a single feed). Subsequent PRs will deal with source rules (i.e. rules that can apply to any feed, but only to items from a particular source, i.e. Articles, Blogs etc.) and general rules (rules that can be applied to any feed that act on all items regardless of their source).

It's probably best just to give Mark or I a shout if further clarification is required.

PLEASE NOTE: UI is in a state of flux and will be changing. Comments still welcome though.


benbarnett:
"PLEASE NOTE: UI is in a state of flux and will be changing. Comments still welcome though."

Will the UI be changing in this PR? If so should we mark this as [WIP]?

urfolomeus:
Sorry @benbarnett no, no more UI changes in this PR. We need to get the other rules in first so we can see how it all fits together and then do more work on the UI. If there are comments on the UI though it would be good to catch them now. Was more just pointing out that what we had was not finished product.

app/views/admin/rules/edit.html.erb
```diff
@@ -0,0 +1,41 @@
...
+    { text: @feed.name, url: admin_partner_feed_url(@partner, @feed) },
+    { text: 'Edit applied local rules' }
+  ] %>
+
+<h1>Applied local rules</h1>
+
+<%= link_to 'Add a local rule', new_admin_partner_feed_local_rule_path(@partner, @feed) %>
+
+<%= form_tag admin_partner_feed_rules_path(@partner, @feed), method: :patch do |form| %>
+  <table>
```
benbarnett: Remember we have the scrollable table: https://www.moneyadviceservice.org.uk/en/styleguide/components/common#section-scrollable-table

urfolomeus: Thanks @benbarnett, will do.

app/views/admin/feeds/show.html.erb
```diff
@@ -11,30 +11,15 @@
...
-        <td><%= rule.name %></td>
-        <td><%= link_to 'Edit', edit_admin_partner_feed_local_rule_path(@partner, @feed, rule) %></td>
-        <td><%= link_to 'Remove', admin_partner_feed_local_rule_path(@partner, @feed, rule), method: :delete %></td>
-      </tr>
-    <% end %>
-  </tbody>
-</table>
+<%= link_to 'manage rules', edit_admin_partner_feed_rules_path(@partner, @feed) %>
+<ul>
+  <li>Pixel tracking <%= @feed.pixel_tracked? ? '✓' : '✗' %></li>
```
tomas-stefano: Can we move this to a helper or a presenter?

app/views/admin/feeds/show.html.erb
```diff
@@ -11,30 +11,15 @@
...
+
+  <% @feed.local_rules.each do |local_rule| %>
+    <% is_applied = local_rule.applied_feed_rules.where(feed_id: @feed.id).exists? %>
```
tomas-stefano: Is it ok to query so many times the database?

app/views/admin/rules/edit.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <table class="datatable-default">
+      <thead>
+        <tr>
+          <th>Rule</th>
+          <th colspan="2">Actions</th>
+        </tr>
+      </thead>
+      <tbody>
+        <% @feed.local_rules.each do |local_rule| %>
+          <% is_applied = local_rule.applied_feed_rules.where(feed_id: @feed.id).exists? %>
```
tomas-stefano: The same query. Can we move to a helper on the controller?

app/views/admin/feeds/show.html.erb
urfolomeus: Hi @tomas-stefano, we have a few different types of rule that we need to apply so we're writing this out long hand just now and will refactor to a cleaner solution once we have found the common elements.

urfolomeus:
We are closing this pull request as we think it is to WIP to be easily reviewed.

PR #33, opened 2015-06-10
Local rules are selectable

urfolomeus:
Refactoring to have a separate process for creating/editing local rules and for applying them.

Before this work local rules would have to be deleted to stop them being applied to a feed. However we want to be able to temporarily turn local rules off without having to delete them.


paulanthonywilson:
:+1: seems to hold together.

darokel:
Just one comment but :+1: otherwise

PR #34, opened 2015-06-10
Rename local rules to adhoc rules

mconnell:
This PR renames local rules to be 'ad hoc' rules so that the codebase terminology matches the client's terminology and avoids future confusion.


darokel:
:+1:

paulanthonywilson:
:+1:

PR #35, opened 2015-06-10
Rename global rules to source rules

mconnell:
This PR renames an existing concept of a 'global rule' to a 'source rule'. Similar to PR #34, this is to make terminology more consistent and less confusing.

To summarise the changes made here.
* Renamed a join model SourceRule to SourceRuleSource. This is a has_many join model that associates sources to what we're referring to as a 'source rule'.

* Renamed a boolean on Rule to determine whether a rule is or is not a source rule instead of a global rule. There is scope to replace this implementation, with a dedicated SourceRule class which would solidify the concept of a 'source rule', but we felt this was outside the scope of this naming refactor.

* Introduced an additional validation that requires a 'source rule' to actually have at least one source.


darokel:
:+1:

paulanthonywilson:
:+1:

PR #36, opened 2015-06-11
Select source rules

mconnell:
This PR introduces the UI to be able to select/deselect which source rules should be applied to a given feed.


app/views/admin/rules/edit.html.erb
```diff
@@ -22,8 +51,11 @@
         <% @feed.adhoc_rules.each do |adhoc_rule| %>
           <tr>
             <td>
-              <%= check_box_tag 'applied_rule_ids[]', adhoc_rule.id, @feed.rule_applied?(adhoc_rule) %>
+              <% id = (Time.now.to_f * 10000).to_i %>
```
paulanthonywilson: Wat?

spec/models/feed_spec.rb
```diff
@@ -87,7 +87,38 @@
...
+    subject { FactoryGirl.create(:feed) }
+    let(:rule) { FactoryGirl.create(:rule, :source_rule) }
+    let(:rule_source) { rule.sources.first }
+    let(:other_source) { FactoryGirl.build(:source) }
+
+    describe 'given a valid source and a valid rule' do
+      context 'when the rule is not applied to the feed' do
+        it 'returns false' do
+          expect(subject.rule_applied?(rule)).to eql(false)
+          expect(subject.rule_applied_for_source?(rule, rule_source)).to eql(false)
```
paulanthonywilson: be_falsey / be_truthy??

app/views/admin/rules/edit.html.erb
urfolomeus: We need to have a unique id for each item but, as it isn't created yet we don't have one. Using a timestamp instead of an incrementing int makes it a bit more obvious that this is a temporary value.

paulanthonywilson: I wonder if a helper named something like ‘temporary_id_for_item’ would make the intent clearer.

mconnell:
@paulanthonywilson made requested changes.

PR #37, opened 2015-06-11
only apply source rules to the corresponding source items

mconnell:
This PR adds the functionality to only apply rules associated one or more sources, and only apply the rule to the source items (blog posts, articles, etc) that have been fetched from the source.


spec/models/applied_feed_rule_spec.rb
```diff
@@ -28,4 +28,39 @@
...
+    let(:source_id) { source_item.source_id }
+
+    context 'when a source rule' do
+      let(:is_source_rule) { true }
+
+      context 'and the source matches the source for the given source item' do
+        let(:source_id) { source_item.source_id }
+
+        it 'is true' do
+          expect(applied_feed_rule.applies_to?(source_item)).to eql(true)
```
paulanthonywilson: Would be_truthy (and later on) be_falsey be more idiomatic and flexible here. Do you care that it's equal to true, or that it's truthy?

spec/models/applied_feed_rule_spec.rb
```diff
@@ -28,4 +28,39 @@
...
+    end
+
+    context 'when not a source rule' do
+      let(:is_source_rule) { false }
+
+      it 'is true' do
+        expect(applied_feed_rule.applies_to?(source_item)).to eql(true)
+      end
+    end
+  end
```
paulanthonywilson: Apart from that, I think this spec nicely explains the behaviour :-)

paulanthonywilson:
:+1: Just with the eq(true) eq(false) query.

spec/models/applied_feed_rule_spec.rb
urfolomeus: Yes, we deliberately choose `eq(true)` here so that we weren't just getting a truthy result but the actual value `true`.

mconnell:
truthy falsey change made.

mconnell:
@paulanthonywilson ^^

paulanthonywilson:
Rethumbing :+1:

PR #38, opened 2015-06-12
Source rules ui

mconnell:
This PR adds the ability to select/deselect source rules for a feed. And the additional logic required for these source rules to be applied to an outputted RSS feed.


tomas-stefano:
:+1:

spec/models/rule_spec.rb
```diff
@@ -166,4 +166,29 @@
...
+        [
+          FactoryGirl.create(:source, name: 'First Source'),
+          FactoryGirl.create(:source, name: 'Another Source')
+        ]
+      end
+
+      subject { FactoryGirl.create(:rule, source_rule: true, name: 'Source Rule', sources: sources) }
+
+      it 'includes source names' do
+        expect(subject.name).to eq('Source Rule (First Source, Another Source)')
```
paulanthonywilson: Nice!

paulanthonywilson:
:+1:

app/models/rule.rb
```diff
@@ -24,6 +24,12 @@ def apply_to(target)
     raise UnknownRuleType
   end
 
+  def name
```
asmega: Is this not polluting the original intention of the method?
Would it not be better to have another method called `name_with_sources` on top of the original method
As developer I would find this frustrating if I just wanted the name of the object only to find some other data injected in.

PR #39, opened 2015-06-12
Changing Source rules index to System rules index

urfolomeus:
This is so that we have one place to change both Source rules and Template rules (once we have them).
Collectively these rules are known as system rules as they happen at a system level and not at an
individual feed level.


tomas-stefano:
:+1:

asmega:
:+1:

PR #40, opened 2015-06-12
Changing rule #name override method to stand alone #label method

urfolomeus:
After a suggestion by @asmega we've refactored this so that we still have access to the #name method


neoeno:
Not sure of the rationale, but it all looks good to me :+1:

tomas-stefano:
:+1:

urfolomeus:
@neoeno it was pointed out by @asmega that you might want to use just name in some cases (which turned out to be the case).

PR #41, opened 2015-06-15
Add feed flow

mconnell:
This PR changes the initial flow of creating a new feed to steer the user through the process of selecting content after creating a new feed.


app/views/admin/firehose_feed_sources/new.html.erb
```diff
@@ -0,0 +1,19 @@
...
+             builder: Dough::Forms::Builders::Validation do |form| %>
+
+  <%= form.collection_check_boxes :source_ids, @sources, :id, :name do |b| %>
+    <%= form.form_row(:source) do %>
+      <%= b.check_box %>
+      <%= b.label { b.text } %>
+    <% end %>
+  <% end %>
+
+  <div>
```
benbarnett: not sure what value this `div` brings?

app/views/admin/partners/show.html.erb
```diff
@@ -19,11 +19,23 @@ Url: <%= link_to @partner.url, @partner.url %>
...
+      <% @partner.feeds.each do |feed| %>
+        <tr>
+          <td><%= feed.name %></td>
+          <td><%= feed.description %></td>
+          <td><%= link_to 'Show feed', admin_partner_feed_path(@partner, feed) %></td>
+        </tr>
+      <% end %>
+    </tbody>
+  </table>
+</div>
```
benbarnett: Newline....

app/views/admin/partners/show.html.erb
```diff
@@ -19,11 +19,23 @@ Url: <%= link_to @partner.url, @partner.url %>
...
-    <%= link_to 'Show feed', admin_partner_feed_path(@partner, feed) %>
-  </div>
-<% end %>
+<div class="table-wrapper">
+  <table class="datatable-default">
+    <thead>
+      <tr>
+        <th>Name</th>
+        <th>Description</th>
+        <th colspan="2">Actions</th>
```
benbarnett: Do we need colspan? Looks like there's only 3 columns below too.

app/views/admin/feeds/_form.html.erb
```diff
@@ -31,7 +31,11 @@
   <% end %>
 
   <%= form.form_row do %>
-    <%= link_to 'cancel', feed.persisted? ? url : [:admin, partner] %>
-    <%= form.submit feed.persisted? ? 'Update feed' : 'Create feed' %>
+    <% if feed.persisted? %>
+      <%= link_to 'cancel', url %>
+      <%=  form.submit 'Update feed' %>
```
benbarnett: Double space.

app/views/admin/feeds/_form.html.erb
```diff
@@ -31,7 +31,11 @@
...
+    <% else %>
+      <%=  form.submit 'Next' %>
```
benbarnett: Double space.

mconnell:
Thanks @benbarnett. Updated accordingly.

benbarnett:
Cool, frontend seems good now :+1:

JiggyPete:
The only thing I can pick up on is that the text in the html files isn't coming from a I18n YAML file?

mconnell:
If we left them inline, we would have started to have multiple places where we had conditional logic appearing in the view. We originally had them split and decided it was easier to read splitting them back out.

mconnell:
We don't require localisation, it's an internal project that only has english users.

JiggyPete:
:+1:

PR #43, opened 2015-06-15
Applying rules for first time

mconnell:
This PR introduces a flow for adding new rules after creating a new feed. The flow after creating a new feed is slightly different to an existing feed, as the user should be presented with all of the source rules check by default, with the option to uncheck them.


app/views/admin/rules/new.html.erb
```diff
@@ -0,0 +1,36 @@
+<h1>Apply Source Rules</h1>
+
+<%= form_tag admin_partner_feed_rules_path(@partner, @feed), method: :post do |form| %>
```
munckymagik: Admittedly, I'm still feeling my way with the Dough stuff, should this be a Dough form builder thingey?

munckymagik:
:+1:

app/views/admin/rules/new.html.erb
Tr4pSt3R: <div id="calc">
  <%= form_for @calculator, url: { action: "create" }, html: { class: 'form' }, builder: Dough::Forms::Builders::Validation do |f| %>
    <%= f.validation_summary %>
.
.
.
</div>

Yes, you may need something like the above to make forms Doughey. 

Example file:
(https://github.com/moneyadviceservice/cutback-calculator/blob/master/app/views/cutback_calculator/calculators/new.html.erb)

rnalexander:
:+1: :shipit:

mconnell:
@Tr4pSt3R @munckymagik dough'd the forms, thanks.

munckymagik:
:100: :+1:

PR #44, opened 2015-06-15
add scope to return list of sources for a curated feed

mconnell:
Adds a scope to allow us to fetch all the of the unique sources for a given curated feed.


asmega:
:+1:

rnalexander:
:+1:

PR #46, opened 2015-06-16
only show relevant source rules for a feed.

mconnell:
Feeds only see source rules that apply to sources that they have.


app/models/rule.rb
```diff
@@ -18,6 +18,10 @@ class Rule < ActiveRecord::Base
 
   scope :source_rules, -> { where(source_rule: true).order(created_at: :asc) }
 
+  scope :source_rules_for, lambda { |ids|
+    joins(:source_rule_sources).where(source_rule_sources: { source_id: ids }).distinct
+  }
```
neoeno: You could use a `do...end` block here

neoeno:
:+1:

app/models/rule.rb
urfolomeus: `do...end` doesn't work with lambdas unfortunately. We tried many variations on this, but this is the only one that Rubocop would allow.

neoeno: Syntactically it seems to be proper, but I do remember @munckymagik coming up against some issue with rubocop and multi-line lambdas...

munckymagik: Yeah there are several options when it's multi-line. All looked ugly.

Incidentally, I think the only way to get the `do-end` lambda to work was to wrap it in parentheses. Obviously that looked ugly, but Rubocop would not allow curly-brackets like this for me, so I ended up making declaring it using the class method approach instead.

munckymagik:
:+1:

PR #48, opened 2015-06-17
GA changes

mconnell:
* Previously we were sending the original URL for an article that a pageview had been recorded for. This change alters this url so all pageviews appear to come from 'www.publish2.com' as there are filters and reporting tools that require this to be set.

* We've also added a 'referrer' attribute to the pageview, so it's easier to identify that the pageview came from webfeeds, and not publish2.com.


munckymagik:
:+1:

app/models/analytics.rb
```diff
@@ -28,8 +29,13 @@ def self.url_with_additional_params(options)
       'webfeeds'              => '1'
     }
 
+    url           = publish2_url(options[:source_item_url])
     params_string = params.map { |k, v| "#{k}=#{URI.escape(v)}" }.join('&')
+    "#{url}?#{params_string}"
```
neoeno: It might not be a scenario where someone could gain much, but since this is open source — an attacker could pass the following sets of malicious parameters:

```ruby
Analytics.url_with_additional_params({source_item_id: '1', partner_name: 'george', partner_id: '2', source_item_url: 'http://attacker.com@attacker.com/'})
# => "http://www.publish2.com@attacker.com/?publish2_story_id=1&publish2_byline=None&publish2_partner_name=george&publish2_partner_id=2&webfeeds=1"
```

Or

```ruby
Analytics.url_with_additional_params({source_item_id: '1', partner_name: 'george', partner_id: '2', source_item_url: 'http://google.com@google.com/broken_redirect?go_to=attacker.com#'})
# => "http://www.publish2.com@google.com/broken_redirect?go_to=attacker.com#?publish2_story_id=1&publish2_byline=None&publish2_partner_name=george&publish2_partner_id=2&webfeeds=1"
```

This could possibly be used for spear phishing, along the lines of the 'faked referrer' stuff you see on GA regularly.

The both issues could be resolved by replacing this with something like:

```ruby
    params_string = params.map { |k, v| "#{k}=#{URI.escape(v)}" }.join('&')
    url = URI(options[:source_item_url])
    URI::HTTP.build(
      host: 'www.publish2.com',
      path: url.path,
      query: params_string,
      fragment: url.fragment).to_s
```

mconnell:
@neoeno whatcha think?

neoeno:
:+1: Looks good. I wasn't sure if you'd want the `fragment` in the URL too — but no reason not to!

PR #49, opened 2015-06-17
Source rule list helper

urfolomeus:
This extracts the displaying of a Source Rule's list of Sources into a helper.

For some reason the rebase with current master caused commit ab9ae3a to show again even though it's already been approved on PR #47. Can be ignored on this PR. Apologies for that.

PR #51, opened 2015-06-18
Source rule helper

mconnell:
Adds a helper method for displaying a list of sources that are associated with a source rule.

This PR replaces PR #49 and corrects a history issue.


app/views/admin/rules/new.html.erb
```diff
@@ -20,19 +20,9 @@
...
-            </td>
-            <td>
-              <ul class="table__cell-list">
-                <% source_rule.sources.each do |source| %>
-                  <li><%= source.name %></li>
-                <% end %>
-              </ul>
-            </td>
+            <td><%= check_box_tag 'applied_rule_ids[]', source_rule.id, true %></td>
+            <td><%= label_tag '', source_rule.name %></td>
```
benbarnett: #lazyquestion Will this end up with the `for="id"` attribute for the checkbox above?

mconnell: yerp

benbarnett:
:+1:

munckymagik:
:+1:

PR #52, opened 2015-06-18
Output as json

mconnell:
Adds the option to output a feed as JSON instead of RSS.

Output follows format from Publish2.


app/views/feeds/show.json.erb
```diff
@@ -0,0 +1,11 @@
...
+  "total_feed_items": <%= @feed.source_items.count %>,
+  "custom_feed_id": <%= @feed.id %>,
+  "timezone": "Europe\/London",
+  "title": "<%= @feed.name %>",
+  "description": "<%= @feed.description %>",
+  "feed_type": "<%= @feed.class.to_s.underscore %>",
+  "newsorg_url":"https:\/\/www.moneyadviceservice.org.uk\/",
+  "newsorg_title":"Money Advice Service",
+  "items": <%= @feed.source_items.map { |source_item| SourceItemPresenter.new(@feed, source_item).to_hash }.to_json.html_safe %>
+}
```
neoeno: Is there an assumption here that titles/descriptions/etc don't have double quote characters in? And if so, is that assumption validated or ensured anywhere?

Also, on lines 4 and 8 — is that `\/` an escaped `/`? If so, why do you need to escape it? And if there's a good reason, are `/` characters escaped in all the other values?

app/presenters/source_item_presenter.rb
```diff
@@ -22,6 +45,33 @@ def to_xml
 
   private
 
+  def datetime_attributes(name, attribute)
+    {
+      name.to_s           => @source_item.send(attribute).iso8601,
+      "#{name}_str"       => @source_item.send(attribute).strftime('%Y%m%d%H%M%S'),
```
neoeno: Might be nice to have this date format in a constant, just to clarify its meaning and purpose

app/presenters/source_item_presenter.rb
```diff
@@ -22,6 +45,33 @@ def to_xml
...
+      'publisher_id'   => @feed.partner_id
+    }
+  end
+
+  def miscellaneous_attributes
+    {
+      'content_status'    => 'editorial',
+      'content_type'      => 'story',
+      'organization_name' => ORGANISATION_NAME,
+      'organization_url'  => 'https://www.moneyadviceservice.org.uk/',
```
neoeno: If `ORGANISATION_NAME` is a constant, should the URL be as well?

neoeno:
One query to do with the JSON generation, and a couple of constant suggestions — but all looks good other than that.

app/presenters/source_item_presenter.rb
urfolomeus: It's only used the once, so probably not worth it at the moment.

app/views/feeds/show.json.erb
urfolomeus: Good points. Will review.

urfolomeus:
@neoeno I believe we may have responded to all of your points now?

neoeno:
:+1:

asmega:
:+1:

PR #53, opened 2015-06-23
Output as js widget

urfolomeus:
This PR allows us to give clients a script tag which then downloads and inserts a feed as HTML. We initially were working towards a script that used AJAX to pull the existing JSON feed but then realised that we would have to also package a sane way of doing AJAX and then a templating function and/or library and ended up figuring that an easier option would be just to use js.erb because

1) we don't have to package any other libraries, vanilla js will do
2) we don't have to worry about getting or holding data, we just send the HTML as a lump with the script


app/views/feeds/_feed.html.erb
```diff
@@ -0,0 +1,11 @@
+<dl class="publish2_stories">
+  <% feed.source_items.each do |source_item| %>
+    <% presenter = SourceItemPresenter.new(feed, source_item) %>
+    <dt>
+      <h2 class="publish2_story_headline"><%= presenter.title %></h2>
+    </dt>
+    <dd style="margin-left: 0px;">
```
benbarnett: Inline style?

urfolomeus: This is going to be displayed on the client's site and so we're not sending CSS. We had to mimic the existing Publish 2 markup so that client's are unaware of the change and so we left the inline style in.

urfolomeus: From discussion on Slack:

"TL;DR the markup that we have used in [this] view is not going to be shown on a MAS site, it will be on a 3rd party partner site (i.e. the consumer of the feed), and there is a specific requirement not to deviate from the existing markup."

JiggyPete:
:+1:

urfolomeus:
As Edd may give more than one feed to a partner and we can't control what they do with those script tags, we thought it prudent to ensure that the ID was unique to a feed. Just in case. :)

urfolomeus:
:+1: on @benbarnett behalf (and at his request) :)

PR #54, opened 2015-06-24
Bug fixen regexp multiline ignore case

urfolomeus:
Fixes two scenarios with Rexexp matching that we missed initially:

1) Matching content dispersed over multiple lines.
2) Using case insensitive matching.


neoeno:
:+1:

rnalexander:
:+1: :shipit:

PR #55, opened 2015-06-25
Email notifications

urfolomeus:
We run a nightly task to pull any new or updated items from the various feed sources. If any items that are attached to a partner's feed have been updated then we need to send them a notification email with the details. We should only send one email per partner per run.

To explain a design decision, we have created two new types of object to help with this: a query object that lets us build an efficient query for gathering the data (using associations would have led to circular relationships and a probable 2n+1 query), and a report object that lets us turn the resulting data from the query into something we can use in the email.

This is currently WIP because we need to improve the UI for adding the emails of those that should be notified and we also need to get the actual copy for the notification email from Edmund. Please note though that Edmund did specifically ask for a free text field to add a comma delimited list of email addresses, as the email may go to people other than the contacts for that partner.


app/reports/updated_source_items_report.rb
```diff
@@ -0,0 +1,17 @@
+class UpdatedSourceItemsReport
+  def for_partner(partner)
+    query_string = UpdatedSourceItemQuery.new.query_string_for_partner(partner)
+
+    result = ActiveRecord::Base.connection.execute(query_string)
```
asmega: why did you choose a handcrafted query over using the model directly?

urfolomeus: I mention it in the commit message for the PR "using associations would have led to circular relationships and a probable 2n+1 query". It's due to the complex interaction between Partner -> two types of Feed (CuratedFeed and FirehoseFeed) -> SourceItems.

Also, we felt that we were giving too much responsibility to the Partner.

app/queries/updated_source_item_query.rb
```diff
@@ -0,0 +1,36 @@
...
+        .on(feed_table[:id].eq(feed_source_item_table[:feed_id]))
+      .join(partner_table)
+        .on(partner_table[:id].eq(feed_table[:partner_id]))
+      .where(
+        (partner_table[:id].eq(partner.id))
+        .and(
+          partner_table[:last_notified_at].lt(source_item_table[:published_at])
+        )
+      ).to_sql
+  end
```
neoeno: So, you clearly spent a lot of time working this out so maybe you have a good reason for doing it this way, but I played around with it for a while and came up with this:

```ruby
class UpdatedSourceItemQuery
  def query_string_for_partner(partner)
    SourceItem.joins(feed_source_items: [feed: [:partner]])
              .where('partners.last_notified_at < source_items.published_at')
              .where('partners.id' => partner.id)
              .select('source_items.title', 'feeds.name').to_sql
  end
end
```

This passes all the tests save one — the one that compares the exact SQL. I've formatted the difference for your convenience:

```SQL
## Arel

  SELECT "source_items"."title", "feeds"."name" 
    FROM "source_items" 
      INNER JOIN "feed_source_items" ON "feed_source_items"."source_item_id" = "source_items"."id" 
      INNER JOIN "feeds" ON "feeds"."id" = "feed_source_items"."feed_id" 
      INNER JOIN "partners" ON "partners"."id" = "feeds"."partner_id" 
    WHERE 
      "partners"."id" = 123 
      AND "partners"."last_notified_at" < "source_items"."published_at"

## ActiveRecord

  SELECT source_items.title, feeds.name 
    FROM "source_items" 
      INNER JOIN "feed_source_items" ON "feed_source_items"."source_item_id" = "source_items"."id" 
      INNER JOIN "feeds" ON "feeds"."id" = "feed_source_items"."feed_id" 
      INNER JOIN "partners" ON "partners"."id" = "feeds"."partner_id" 
    WHERE 
      (partners.last_notified_at < source_items.published_at) 
      AND "partners"."id" = 123
```

Either way, it could do with a comment that explains what you're trying to end up with here. E.g. 'We want a list of source items that are being used by a feed' (or whatever it is)

urfolomeus: Ha! Yeah that's infinitely more readable. Thanks for that. Was definitely one of those "lost in the weeds" moments. :)

Also agree on the comment comment :)

urfolomeus: Might be worth us considering losing this query object and putting this method onto SourceItem seeing as this gets rid of the private methods and the need to turn off Rubocop for readability. Will discuss with @mconnell in the morning (after he's finished punching me in the head). ;)

app/reports/updated_source_items_report.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    source_item_data = {}
+
+    result.each do |result_item|
+      source_item_title = result_item['title']
+      source_item_data[source_item_title] ||= []
+      source_item_data[source_item_title] << result_item['name']
+    end
```
neoeno: I kind of think this is worse, but here's an alternative in case you disagree:

```ruby
result.group_by { |item| item['title'] }.map do |title, results|
  [title, results.map { |r| r['name'] }]
end.to_h
```

db/migrate/20150624122118_add_last_notified_at_to_partners.rb
```diff
@@ -0,0 +1,13 @@
+class AddLastNotifiedAtToPartners < ActiveRecord::Migration
+  class Partner < ActiveRecord::Base; end
+
+  def change
+    add_column :partners, :last_notified_at, :datetime
+
+    # ensure that partners have a last_notified_at date
+    now  = Time.zone.now
```
neoeno: Doppelspace here

app/reports/updated_source_items_report.rb
urfolomeus: Personally I find this harder to parse but let's put it to a vote. :)

db/migrate/20150624122118_add_last_notified_at_to_partners.rb
urfolomeus: Good catch.

neoeno:
Ruby-wise, seems great.

I've noticed one of the tests fails if I don't have `CONTACT_EMAIL` set in my env. This was very distressing to me personally and I am contemplating suing you for damages. Maybe you could set a default value for test mode? Or in the `Rakefile`?

urfolomeus:
You Millenials and your sense of entitlement :P

Yeah good shout. Will add to README and set a default in rails_helper or the like.

Tr4pSt3R:
Hey Pruner, I find it hard to follow the breakdown of the scenarios without a Gherkin description or Cucumber scenarios. They do certainly make it easier to do reviews.

urfolomeus:
Hi @Tr4pSt3R, not sure I follow?

Tr4pSt3R:
Hey @urfolomeus something like this: "If any items that are attached to a partner's feed have been updated then we need to send them a notification email with the details. We should only send one email per partner per run." 

With this being the acceptance scenario, I expected an acceptance test so that I can follow the story from an outsiders point of view.

urfolomeus:
Hi @Tr4pSt3R, there actually isn't really an outsiders point of view on this one. It is a background task that sends emails. We have tests in place that cover all of logic and even the resulting email. Perhaps what we need is better explanations within our specs?

urfolomeus:
@asmega @neoeno I've made the change on the query object and commented on that commit as to why I did what I did. Does this resolve your concerns?

app/reports/updated_source_items_report.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    source_item_data
+  end
+
+  private
+
+  def report_data_for(partner)
+    ActiveRecord::Base.connection.execute query_string_for_partner(partner)
```
neoeno: From the commit message:

> It still make sense to keep this separate from the ActiveRecord models that it invokes as the returned data is specific to this report and serves no other purpose. 

Does this refer to the process of going to and from SQL rather than executing the query directly?

urfolomeus: Hmm, good point. >.<

urfolomeus:
@neoeno @asmega OK simplified again. I tried moving the bulk of the query to SourceItem as it seemed to belong there now (minus the `.select` that is). However it still feels too bespoke to be used elsewhere (it will return duplicate source_items if they belong to more than one feed, which is what we want for the report but is counter-intuitive to have as a default IMHO). Have left it where it is. Personally I think it makes to be close to the report that uses it when nothing else in the system has a requirement for anything similar at the moment.

neoeno:
I'm with you. Is this still WIP? Or is it ready for some sort of hand emoji?

urfolomeus:
@neoeno @asmega @Tr4pSt3R I've taken WIP off now as we'll be covering the change of email copy and additional UI requirements under a separate PR. The email copy changes will be minimal according to Edmund and I don't want to change the UI too much just now as @spencer and @mconnell are doing a separate piece of work to improve the UI just now.

Have at it!

neoeno:
:+1:

README.md
```diff
@@ -48,11 +55,21 @@ cap production deploy
 You may also want/need to run `rake db:seed` on the server.
 
 ### ENV variables
-The following are sample ENV variables that need set in a production environment:
+The following are the ENV variables that need set in
+
+* development:
+```bash
+  SKIP_SLOW_TESTS=(true|false)
+  CONTACT_EMAIL=partners@moneyadviceservice.org.uk
```
asmega: why does this have to be an environment variable? can't it be a constant/method in a relevant class?

db/migrate/20150624122118_add_last_notified_at_to_partners.rb
```diff
@@ -0,0 +1,13 @@
+class AddLastNotifiedAtToPartners < ActiveRecord::Migration
+  class Partner < ActiveRecord::Base; end
+
+  def change
+    add_column :partners, :last_notified_at, :datetime
+
+    # ensure that partners have a last_notified_at date
+    now = Time.zone.now
+    Partner.find_each do |partner|
+      partner.update_attribute(:last_notified_at, now)
```
asmega: fyi http://apidock.com/rails/ActiveRecord/Persistence/touch

README.md
urfolomeus: We'd like to be able to set it on a per environment basis for debugging purposes.

db/migrate/20150624122118_add_last_notified_at_to_partners.rb
urfolomeus: Ah yeah, thanks. We did use that in the job. @mconnell introduced me to it yesterday :)

app/models/partner.rb
```diff
@@ -5,4 +5,12 @@ class Partner < ActiveRecord::Base
   accepts_nested_attributes_for :contacts, reject_if: :all_blank, allow_destroy: true
 
   validates :name, :url, presence: true
+
+  before_create :set_last_notified_at
+
+  private
+
+  def set_last_notified_at
```
asmega: should this not be a callback on the actionmailer?

urfolomeus: This is the `last_notified_at` being set initially when the Partner is created so that they don't get spammed the day after we created their first feeds with everything that we added to them. The updating thereafter happens as part of the `NotifyPartners` job.

README.md
asmega: In which I suggest something like

some module or class. could be a method or constant
```
def contact email
   ENV['CONTACT_EMAIL'] || 'partners@moneyadviceservice.org.uk'
end
```
now this belongs somewhere meaningful and appropriate you can call this method and you won't have `ENV['CONTACT_EMAIL']` everywhere

app/models/partner.rb
asmega: :+1:

README.md
urfolomeus: I'm not sure it makes sense to or with the literal version as that leaves two places to edit it if changes, but certainly happy to wrap it in a method.

asmega: sure. method is good

urfolomeus: @asmega on second thoughts it's as easy to change a constant as it is to change an environment variable. Changed to a constant. :)

asmega:
:+1:

urfolomeus:
@asmega moved it inline as you suggested. Thanks.

PR #59, opened 2015-06-30
Optionally remove headlines from feed items on the JS Widget

urfolomeus:
Some partners use the JS Widget within a CMS that already has a title field
that is used to put in the headline of the feed item. As such we can optionally
serve that content without the headlines.


app/helpers/feed_helper.rb
```diff
@@ -1,6 +1,6 @@
 module FeedHelper
-  def script_tag_for(feed)
-    source = widget_feed_url(feed, protocol: '//', format: :js)
+  def script_tag_for(feed, exclude_headlines)
```
asmega: I'd use an options hash here so the api is more flexible to change

urfolomeus: A good point. I'll make that change. Thanks.

asmega: or named argument

munckymagik:
:+1:

urfolomeus:
@asmega implemented as suggested.

asmega:
:+1:

PR #60, opened 2015-06-30
Partners are ordered by name

urfolomeus:


asmega:
:+1:

urfolomeus:
:point_up: This was just rebasing against master after merging PR #59.

app/views/admin/partners/index.html.erb
```diff
@@ -12,7 +12,7 @@
       </tr>
     </thead>
     <tbody>
-    <% @partners.all.each do |partner| %>
+    <% @partners.each do |partner| %>
```
munckymagik: Deliberately not using the new scope here?

munckymagik: Aha because it's in the controller. Ok ignore me.

spec/models/partner_spec.rb
```diff
@@ -19,6 +19,17 @@
...
+    describe 'by_name' do
+      let!(:zulu) { FactoryGirl.create(:partner, name: 'Zulu') }
+      let!(:acme) { FactoryGirl.create(:partner, name: 'Acme') }
+
+      it 'orders the partners by name' do
+        expect(Partner.by_name).to match_array([acme, zulu])
+      end
+    end
+  end
+
```
munckymagik: If you're using Postgres this may not be enough. We have a similar test that randomly ended up giving us back ordered result even though we hadn't explicitly ordered them, therefore breaking the test. Paul's suggestion was using more records (e.g. 8). See: https://github.com/moneyadviceservice/mas-rad_core/pull/74/files

urfolomeus: Good shout, will do this just now. Thanks :)

munckymagik:
:+1:

PR #61, opened 2015-07-01
Header styling

urfolomeus:
This is the first of a few PRs for UX and styling work on the project. This PR deals with the styling of the header for the site.


app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,36 @@
-.header__settings {
+header h1 { font-size: 0.5em;}
```
Guntrisoft: BEM?

benbarnett: and @font-size mixin

app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,36 @@
...
+}
+
+.header__item--last {
+  border-right: 0;
+}
 
-  &:visited {
-    color: $color-yellow-light;
+.header__link,
+.header__item > a {
```
benbarnett: BEM naming again

app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,36 @@
...
+  @extend .unstyled-button;
+  @include body(14, 14);
+  color: white;
```
benbarnett: Should be a colour variable

app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,36 @@
...
+
+  &:hover,
+  &:focus {
+    color: white;
```
benbarnett: colour variable

app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,36 @@
...
+    outline: none;
```
benbarnett: Why outline:none out of interest? This is, in general, bad for accessibility.

app/assets/stylesheets/_layout.scss
```diff
@@ -13,6 +13,8 @@
     @include column(12);
     padding: $baseline-unit*4 0;
   }
+
+  .l-1col { background: none;}
```
benbarnett: Please always use drop the statements on to a new line, so block style rather than inline

app/assets/stylesheets/_header.scss
spencer: This was taken from authentication styles in front-end as it was visually what we needed, so I think it came in from there.

benbarnett: Looks like it originates for when it was [on a submit button](https://github.com/moneyadviceservice/frontend/commit/db8195e1cccf0f6215b84a9b73487d76ba59e07b), but as you're just using anchor links this should be removed. We'll fix up frontend too.

benbarnett: Ultimately though, as long as there's a visual state for hover & focus, then this is ok.

urfolomeus:
@benbarnett @Guntrisoft I believe we've addressed your feedback. Could you review, please? Thanks.

app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,34 @@
...
+}
+
+.header-settings__item--last {
+  border-right: 0;
+}
 
-  &:visited {
-    color: $color-yellow-light;
+.header-settings__link,
+.header-settings__item > a {
```
Guntrisoft: Is this not the same thing, twice?

the links are given the class .header-settings__link, so no need for the second statement?

app/assets/stylesheets/_header.scss
benbarnett: We appear to have lost this styling altogether now – was that intentional?

app/assets/stylesheets/_header.scss
spencer: Again, direct copy and paste (with name adjustment) from frontend, looked like 'belt and braces' to cover both cases, so left it.

app/assets/stylesheets/_header.scss
urfolomeus: Yes. It's not actually required.

app/assets/stylesheets/_header.scss
benbarnett: Still probably worth cleaning it up here, its confusing to read.

app/assets/stylesheets/_header.scss
```diff
@@ -1,12 +1,37 @@
...
+  @include body(14, 14);
+  color: $color-white;
+
+  &:hover,
+  &:focus {
+    color: $color-white;
+    text-decoration: underline;
+  }
 
   &:visited {
```
urfolomeus: It would appear that the `.header-settings__item > a` was superseding a general visited link style, so I had to add this too.

urfolomeus:
For reference:

# Old header style



# New header style

urfolomeus:
@benbarnett @Guntrisoft have we addressed all concerns now or is there still work to do on this PR? Thanks.

urfolomeus:
Rebased to master after swapping to RAD way of pulling in MAS styles.

benbarnett:
Would be interested to see how much of Dough we can remove now that we're pulling in the 'generated' styles. But this component seems good now :+1:

urfolomeus:
That's a good point. Will strip out on a separate PR so as not to confuse this one.

urfolomeus:
Slimmed down after rebase.

spencer:
:+1:

PR #63, opened 2015-07-01
Typography styling

urfolomeus:
Updating the typography mostly involved copying and pasting files from frontend and hooking them up. The main purpose was to pull in the existing MAS typography styling. We appreciate that there's still work to be done around spacing and the like.

# Old style

https://www.dropbox.com/s/xvh6bkw3nlnqtxv/Screenshot%202015-07-01%2019.26.52.png

# New style

https://www.dropbox.com/s/onc12cubkcuomsr/Screenshot%202015-07-01%2019.25.53.png


app/assets/stylesheets/_enhanced.scss
```diff
@@ -1,29 +1,35 @@
...
 @import 'table';
 @import 'button';
 @import 'breadcrumb';
 @import 'header';
+@import 'typography';
 
-body {background: #DDD;}
+body {
+  background: #DDD;
+}
```
urfolomeus: It's probably worth pointing out that an imminent PR changes this to a color variable and moves all these styles to a more suitable file.

app/assets/stylesheets/base/_variables.scss
```diff
@@ -24,3 +26,16 @@ $color-button-primary-disabled: #f2ebb2;
...
+
+// Override Yeast variables
+
+$color-heading-default: $color-true-black;
+$color-link-default: $color-blue-cerulean;
+
+
+// New colors
+
+$color-yellow-dark: #e8b940;
```
Guntrisoft: We used this tool (http://chir.ag/projects/name-that-color/) in the past to give the colour a 'proper' name, rather than trying to describe the colour which often ends in problems down the line.

urfolomeus: This color variable is lifted directly from the MAS Styleguide https://www.moneyadviceservice.org.uk/en/styleguide/css

urfolomeus: @Guntrisoft in fact that whole file is lifted wholesale from frontend :)

urfolomeus: @Guntrisoft actually my bad. The name is lifted but the color is changed. I'm not sure why that was done but I'm going to revert it as it seems like bad practice to me.

urfolomeus: OK, this is confusing. It looks like it was added manually (and accidentally with the wrong value) because we're not pulling it in from Dough/Yeast/MAS Dough Theme (or I've somehow not hooked things up correctly).

urfolomeus:
@Guntrisoft after much confusion I've just removed those color variables :)

app/assets/stylesheets/_typography.scss
```diff
@@ -0,0 +1,12 @@
...
+}
+
+.with-button {
+  margin-top: 0;
+  margin-right: 2em;
+  float: left;
+}
+
+// Work around floated h1
+p {clear: both}
```
Guntrisoft: brackets on seperate lines - seems rather generic for all Ps to get this?

spencer: As this is a local style for the admin interface only, it's not problematic.

benbarnett: It is a code smell though :)

benbarnett:
I'm not sure the new styles look quite right, the headings are bumping right up against their previous paragraphs. Is this intentional?

app/assets/stylesheets/_typography.scss
urfolomeus: Just to clarify, you'd prefer it if we put this style on a class and added it to the `<p>` tags on the site instead?

benbarnett: I'd prefer it if we didn't have to clear:both; on every single paragraph. Its indicative of some floats going wrong somewhere and potentially a flaw in the CSS design.

urfolomeus:
@benbarnett this is fixed by PR#61 but we can't see it in here I'm afraid until we merge that in.

benbarnett:
@urfolomeus but zeroing out the margin-top on *all* headings means you'll be continuously be adding it back in? #61 might fix that specific use case, but I can pretty much guarantee you'll end up writing more styles as a result of having no margins on the h1's. I'm talking about [this line](https://github.com/moneyadviceservice/webfeeds/pull/63/files#diff-956b372e3d32f4879e0e77e8cf8daa3cR1) just so we're clear.

I'm going to pick my battles though. If you guys are happy with it, then I'm happy to let it go. :+1:

urfolomeus:
@benbarnett tbh I think that we want to get rid of as much of our custom styling as possible. I'm not sure whether we should do this now on this PR or whether we this until we do PR#64 in which we pull in the dough theme stuff from frontend (as proposed by @Guntrisoft), which should hopefully negate the need for our additional styling?

urfolomeus:
@benbarnett @Guntrisoft I've removed the custom typographic styles for just now with the view to fixing them properly when we import in the frontend styles.

urfolomeus:
Slimmed down after rebase.

urfolomeus:
Barely worth having this commit now! :D

benbarnett:
Heh, still useful!

:+1: after squashes and a little history tidy up

urfolomeus:
@benbarnett I was wondering if it might be worthwhile leaving the history as it is for future reference?

benbarnett:
@urfolomeus yeah, fair enough, although not sure about 'Oops forgot to delete unrequired files' – perhaps a quick reword?

urfolomeus:
Um, yeah that one's not particularly informative :blush:

benbarnett:
:+1:

spencer:
:+1:

PR #66, opened 2015-07-03
Now that we pull in styles from frontend some of our own are redundant

urfolomeus:


urfolomeus:
@benbarnett would seem we can remove quite a lot :)

benbarnett:
My favourite kind of PR! :+1:

urfolomeus:
Moi aussi! :)

spencer:
:+1:

PR #68, opened 2015-07-06
Standalone pixel tracker

mconnell:
Edmund wanted to be able to create pixel tracking image tags for custom articles that are not available from the standard sources.


urfolomeus:
See it in action

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Requires a component with a wrapping div that has a `data-dough-component="PixelTrackerGenerator"` attribute.
+ * The application file will spawn an instance of this class for each element it finds on the page.
+ *
+ * See test fixture for sample markup - /spec/js/fixtures/PixelTrackerGenerator.html
+ */
+
+/**
+ * Require from Config
+ * @param  {[type]} $                  [description]
```
munckymagik: Are the `[type]` and `[description]` bit meant to be like that?

spec/javascripts/fixtures/PixelTrackerGenerator.html
```diff
@@ -0,0 +1,22 @@
+<h2>Generate tracking pixel</h2>
+
+<div data-dough-component="PixelTrackerGenerator">
+  <form class="form" action="test/tracker/url"method="post">
```
munckymagik: Some spacing issues on this line.

spec/javascripts/fixtures/PixelTrackerGenerator.html
```diff
@@ -0,0 +1,22 @@
...
+    <input name="utf8" type="hidden" value="✓">
+    <input type="hidden" name="authenticity_token" value="ovQwMhaxf3HffcuzpbTslFTHDxfFUvwk/up+I6avvqOUsMsy8FXnoOtnz9VQiVvktJka31u8ErCcoD/86aWLaw==">
```
munckymagik: Are lines 5 & 6 used by the JS test? As someone reviewing the test, possibly to understand how the JS component works, would prefer to see the absolute bare bones required to make it work, so preferably no extraneous styles, attributes or form fields. Then you can really see the elements it relies on to work.

app/assets/javascripts/modules/PixelTrackerGenerator.js
urfolomeus: TL;DR I don't know.

We copied this style from another project as that we were following some kind of standard. This isn't documented, as far as I am aware, so we figured this would at least make things consistent.

spec/javascripts/fixtures/PixelTrackerGenerator.html
urfolomeus: I disagree. I want the markup here to represent what will be on the actual site in case some elements affect the tests in some way. If we take out an element because it's not referred to directly in our tests it might have an impact that we wouldn't see if we took it out.

In fact, we're missing some attributes that we added yesterday. So I would argue that we need to add those too.

munckymagik: Well I disagree too :)

However, I think both our opinions seem reasonable, as this is a deliberate approach I'll say good stuff, keep it how it is.

urfolomeus: Happy to be overruled. :)

Personally I hate the fact that this file needs to be manually updated whenever changes are made to the views. It's too easy to forget to update this.

munckymagik:
:+1:

urfolomeus:
With updates the functionality now looks like this ...

app/assets/javascripts/modules/PixelTrackerGenerator.js
benbarnett: This is jsDoc, and should be populated really. Its parsed to produce documentation: https://github.com/moneyadviceservice/dough#javascript-documentation

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,95 @@
...
+   * Handle form submission
+   * @param {object} e
+   * @private
+   */
+  PixelTrackerGeneratorProto._handleFormSubmit = function(e) {
+    var self = this;
+
+    // check if another component eg. Validation, stopped the form submission, before syncing
+    if (e.isDefaultPrevented() === false) {
+      self._doSubmit(e);
```
benbarnett: not sure you need the `self` here?

benbarnett: (i.e. you're still in the same scope so `this` will match)

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,95 @@
...
+
+  PixelTrackerGeneratorProto._doSubmit = function(e) {
+    e.preventDefault();
+
+    var self = this,
+        $form = $(self.$el.find('form')),
+        formValues = {
+          'authenticity_token': $form.find('input[type="hidden"][name="authenticity_token"]').val(),
+          'title': $form.find('input#title').val(),
+          'url': $form.find('input#url').val()
```
benbarnett: Finicky, but these should really be named appropriately `[data-dough-pixel-tracker-title]`, [data-dough-pixel-tracker-url]`, etc

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,95 @@
...
+        };
+
+    $.ajax({
+      method: 'POST',
+      url: $form.attr('action'),
+      dataType: 'json',
+      data: formValues
+    })
+      .done(self.didSucceed)
```
benbarnett: Same here with `self` – you're in the same scope so can just use `this`

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,95 @@
...
+      .fail(self.didFail);
+  };
+
+  PixelTrackerGeneratorProto.didSucceed = function(data) {
+    $('textarea#tracking-pixel').val(data.tracker);
```
benbarnett: This could be given a proper target and cached by the module so you can reference it directly

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  PixelTrackerGeneratorProto.didFail = function() {
+    $('textarea#tracking-pixel').val('An error occurred whilst generating the pixel');
```
benbarnett: Same here re caching the target and naming appropriately

spec/javascripts/fixtures/PixelTrackerGenerator.html
benbarnett: FWIW I agree with @munckymagik magic here. I see the fixtures a description/spec of how this component works in its most generic form.

app/assets/javascripts/modules/PixelTrackerGenerator.js
urfolomeus: @benbarnett Good catch. We had meant to do that but forgot. Not sure we should be doing that on the form inputs though. Will that not mean that the params posted back to Rails will not use the Rails defaults? Apologies if I'm going over old ground here. Happy to make the change if that's house rules. Just checking.

benbarnett: How do you mean the params sorry? ID's are used by the browser, not the server. Or is that not what you mean?

We tend to just provide an additional data attribute to the rails helper, so the `name` stays consistent with what rails wants.

urfolomeus: Sorry I thought you meant to change the name not add a data attribute. My bad. :)

spec/javascripts/fixtures/PixelTrackerGenerator.html
urfolomeus: But what if the other markup has some affect not anticipated? The tests would not be able to catch this if we just have a minimal representation of the markup in question.

benbarnett: Where does it end though? If the markup is having an affect, then the component is being used incorrectly no? Is there a risk of someone new coming to this and assuming that the component somehow requires a `utf8` and `authenticity_token` to function properly?

To avoid batting heads on it, perhaps you could run the tests against two fixtures? One minimal, one as its used in context here?

urfolomeus: Actually no, I concede to your point. By my argument you would have to include the whole page's HTML just in case.

Every day is a school day. :)

urfolomeus:
@munckymagik @benbarnett I think we've addressed concerns now?

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,92 @@
...
+      url: $form.attr('action'),
+      dataType: 'json',
+      data: formValues
+    })
+      .done(this.didSucceed)
+      .fail(this.didFail);
+  };
+
+  PixelTrackerGeneratorProto.didSucceed = function(data) {
+    $('[data-dough-tracking-pixel]').val(data.tracker);
```
benbarnett: Sorry one more thing, this should be scoped to within `this.$el` in the case of multiple of these components per page. I know highly unlikely in this case, but could make for some confusing consequences if not. So `this.$el.find('[data-dough-tracking-pixel]')`

Another optimisation step would be to cache them as an instance variable on init, so you're only scanning the DOM once.

```js
// init
this.$trackingPixel = this.$el.find('[data-dough-tracking-pixel]');

// didSucceed
this.$trackingPixel.val(data.tracker);
```

Completely appreciate this may seem OTT for this module, but just as a best practice thing.

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,92 @@
...
+   * @private
+   */
+  PixelTrackerGeneratorProto._handleFormSubmit = function(e) {
+    // check if another component eg. Validation, stopped the form submission, before syncing
+    if (e.isDefaultPrevented() === false) {
+      this._doSubmit(e);
+    }
+  };
+
+  PixelTrackerGeneratorProto._doSubmit = function(e) {
```
benbarnett: Could also benefit from the jsDoc comments on these methods (and the ones below) too. Not sure what IDE you use but there are typically extensions in most of the popular ones to generate these very easily. Sublime's is here: https://github.com/spadgos/sublime-jsdocs.

benbarnett:
Appreciate I'm being quite specific about these comments, but it helps as a 'practice' module for future reference.

urfolomeus:
I think it's good to have somewhere to check back to. We just need to find somewhere to document that this is the place to look now. :)

app/assets/javascripts/modules/PixelTrackerGenerator.js
urfolomeus: Thanks. I'm using sublime so will add this package and see how I get on.

munckymagik:
Well, SLIM js test fixture! Didn't expect that. Nice. So :+1: for the things I commented on.

urfolomeus:
@benbarnett done as requested

urfolomeus:
Rebased against master after other PRs accepted.

app/assets/javascripts/modules/PixelTrackerGenerator.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+
+/**
+ * Require from Config
+ * @param  {object} $                  jQuery
+ * @param  {object} DoughBaseComponent Dough
+ * @return {object}                    PixelTrackerGenerator
+ * @private
+ */
+define(['jquery', 'DoughBaseComponent', 'Validation'], function($, DoughBaseComponent) {
```
benbarnett: No need for `validation` here as it'll be pulled in by ComponentLoader.

app/views/admin/partners/_generate_tracking_pixel.html.erb
```diff
@@ -0,0 +1,20 @@
+<h2>Generate tracking pixel</h2>
+
+<div data-dough-component="PixelTrackerGenerator">
```
benbarnett: Change this to `<div data-dough-component="PixelTrackerGenerator Validation">` and it should be pulled in for you.

urfolomeus: We actually already have that added to the form and we don't use it outside of the form. Removed from the script though.

benbarnett: Cool cool

urfolomeus:
@benbarnett does this resolve remaining issues now?

benbarnett:
Looks good :+1:

urfolomeus:
Thanks @benbarnett.

PR #69, opened 2015-07-07
Add firehose feeds to the report query

mconnell:
We were missing source items that were attached to Firehose Feeds because source items weren't directly associated. This PR corrects that.


tomas-stefano:
:+1:

neoeno:
:+1:

PR #71, opened 2015-07-08
after saving/updating, redirect the user to the show action

mconnell:
Small PR to change a couple of controller redirects to send the user to the show page rather than the index page.


asmega:
:+1:

urfolomeus:
:+1:

PR #72, opened 2015-07-08
Add url and published at for feed content

mconnell:
Adds a published date and url for a source item into lists & form checkboxes.


asmega:
:+1:

Tr4pSt3R:
:+1:

PR #76, opened 2015-07-09
Poll unpublished

mconnell:
We need to poll available unpublished feeds, and mark any content that is no longer published. This PR covers the rake task to poll the content. Separate rake tasks will then use that data.


urfolomeus:
:+1:

darokel:
:+1:

PR #81, opened 2015-07-10
Add form button wells

spencer:
screen shot 2015-07-10 at 12 17 52


app/assets/stylesheets/components/_forms.scss
```diff
@@ -0,0 +1,21 @@
+.form__label-heading, .form__group-item {
+  padding-bottom: $baseline-unit;
+}
+
+.well {
```
neoeno: This isn't part of the form component, so it should probably be in its own `components/_well.scss` file

app/assets/stylesheets/_enhanced.scss
```diff
@@ -18,6 +18,7 @@
 //Our own stylesheets
 @import 'components/typography';
 @import 'components/button';
+@import 'components/forms';
```
neoeno: The component is `form` so it should be named accordingly.

app/assets/stylesheets/components/_forms.scss
```diff
@@ -0,0 +1,21 @@
...
+  background: $color-green-paler;
+  padding: $baseline-unit *2;
+  text-align: center;
+
+  .button {
```
neoeno: Nesting an existing component within a new component is kind of against the spirit of components. Not sure exactly what this /should/ be doing without more idea of the intent, though.

app/assets/stylesheets/components/_forms.scss
```diff
@@ -0,0 +1,21 @@
...
+    margin-bottom: 0;
+    display: static;
```
neoeno: Is `static` a valid value for `display`?

app/assets/stylesheets/components/_forms.scss
```diff
@@ -0,0 +1,21 @@
...
+    margin: 0 auto;
```
neoeno: Does this overwrite the `margin-bottom` on line 11?

app/views/admin/feed_source_items/edit.html.erb
```diff
@@ -22,8 +22,8 @@
     <% end %>
   <% end %>
 
-  <%= form.form_row do %>
-    <%= link_to 'cancel', admin_partner_feed_path(@partner, @feed) %>
+  <div class="form__row well">
```
neoeno: You could use `<%= form.form_row classes: 'well' do %>` I think

neoeno: Or, actually — it needs to be:

`<%= form.form_row :field_name, classes: 'well' do %>`

neoeno: Maybe

neoeno: Actually, come to think of it, I've never used `form__row` for submit buttons — I've always produced something else. But don't have a source for that...

spencer: we couldn't get this to work. (tried many variants,  but not classes:) will revisit

spencer: tested - can't get it working, hence we went with HTML

app/assets/stylesheets/components/_forms.scss
spencer: fixed

app/views/admin/feed_source_items/edit.html.erb
neoeno: Yeah, sorry, I could have looked it up, the actual usage can be found [here](https://github.com/moneyadviceservice/dough/blob/cf4e4cbd5ee86ef49b2db16fc44dde1ed9f2c0f4/spec/dummy/app/views/forms/_form_row_with_options.html.erb#L4)

Though, in this case, I'd question whether this should be a `form_row` at all — but only because I've not seen them used for submit buttons. frontenders might know better.

app/assets/stylesheets/components/_well.scss
```diff
@@ -0,0 +1,15 @@
+.well {
+  background: $color-green-paler;
+  padding: $baseline-unit *2;
+  text-align: center;
+
+  .button {
+    margin: 0 auto;
```
neoeno: This appears to have no effect — can it be removed?

app/views/admin/adhoc_rules/_form.html.erb
```diff
@@ -30,8 +30,8 @@
     <%= form.text_area :content %>
   <% end %>
 
-  <%= form.form_row do %>
-    <%= link_to 'cancel', edit_admin_partner_feed_rules_path(partner, feed) %>
+  <%= form.form_row(:nil, html_options: {classes: 'well'}) do %>
```
neoeno: We could use `nil` rather than `:nil` here

app/views/admin/sources/_form.html.erb
```diff
@@ -4,7 +4,7 @@
 
   <%= form.validation_summary %>
 
-  <%= form.form_row(:name) do %>
+  <%= form.form_row :name, classes: 'test' do %>
```
asmega: test?

spencer:
Ref: test class, we were trying to find out how to pass a class into the form_row helper, this snuck through.

app/assets/stylesheets/_enhanced.scss
```diff
@@ -18,6 +18,8 @@
 //Our own stylesheets
 @import 'components/typography';
 @import 'components/button';
+@import 'components/form';
+@import 'components/form-action-row';
```
neoeno: So because our 'block' is the same here — the first component of our BEM class is `form` — we can put it all into `components/form` (I think).

app/assets/stylesheets/components/_form-action-row.scss
```diff
@@ -0,0 +1,13 @@
+.form__action-row {
+  background: $color-green-paler;
+  padding: $baseline-unit *2;
+  text-align: center;
+  zoom: 1;
```
neoeno: The clearfix might not be specifically needed here, and if it is and you have the dough sass helpers you could stick `@include clearfix;` at the top here.

app/assets/stylesheets/components/_form-action-row.scss
```diff
@@ -0,0 +1,13 @@
+.form__action-row {
```
neoeno: Ref your query — this is exactly the right BEM as I understand it. You could also probably use `form__row--action` but I personally prefer this — others will have different opinions (and are encouraged to share if they do!)

app/assets/stylesheets/components/_form-action-row.scss
```diff
@@ -0,0 +1,13 @@
...
+  background: $color-green-paler;
+  padding: $baseline-unit *2;
+  text-align: center;
+  zoom: 1;
+  margin-bottom: $baseline-unit *8;
+}
+
+.form__action-row--cancel {
```
neoeno: In this case what we're after is `.form__action-row__cancel`. The reason for this is that our `cancel` here is an element, not a modifier. Use of a modifier might be:

`.form__action-row--highlighted`

A good heuristic here is that if you're in this situation where you're _containing_ a new element:

```html
<div class="block__element">
  <div class="???"></div>
</div>
```

`???` is a sub-element — `block__element__sub-element`.

(Also worth noting that you don't want to take this too far! `block__element__element` is really the absolute farthest you want to go, and some would consider even that a code smell.

app/views/admin/rules/new.html.erb
```diff
@@ -29,7 +29,7 @@
     </table>
   </div>
 
-  <%= form.form_row do %>
+  <div class="form__action-row">
     <%= button_tag 'submit', class: 'button button--primary' %>
```
neoeno: Example of BEM modifier usage in dough/yeast here (though usually we'd want `button--primary` to include all the styles of `button` and so avoid the duplication — not suggesting you change this though!)

app/assets/stylesheets/components/_form-action-row.scss
mconnell: @neoeno just reviewing this at the moment. RE: clearfix. I'm not sure I'm following this comment?

mconnell:
@neoeno addressed other comments other than the clearfix question

app/assets/stylesheets/components/_form.scss
```diff
@@ -0,0 +1,15 @@
+.form__action-row {
+  background: $color-green-paler;
+  padding: $baseline-unit * 2;
+  text-align: center;
+  margin-bottom: $baseline-unit * 8;
+}
+
+.form__action-row__cancel {
```
Guntrisoft: I tend to avoid double nesting of elements in BEM notation so .form__action-row-cancel

app/views/admin/rules/new.html.erb
neoeno: In case anyone sees this — seems like we prefer this style rather than `@extend` for MAS projects. [See this article.](http://www.sitepoint.com/working-bem-scale-advice-top-developers/)

neoeno:
:+1:

benbarnett:
:+1:

PR #83, opened 2015-07-13
Tidy feed show page

mconnell:
This PR adds some layout changes to reduce the vertical height of the feed#show page.

Before:
screenshot 2015-07-13 15 05 31

After:
screenshot 2015-07-13 14 56 09


app/assets/stylesheets/components/_divided.scss
```diff
@@ -0,0 +1,5 @@
+.divided {
```
mconnell: @neoeno wondering if this class should be `.l-divided` as it is changing layout?

app/assets/stylesheets/_enhanced.scss
```diff
@@ -26,6 +26,8 @@
 @import 'components/contact_list';
 @import 'components/blank_state';
 @import 'components/source_item';
+@import 'components/4_col.scss';
```
neoeno: In RAD we put layout classes in `layouts/` rather than `components/` — think that's the standard.

app/assets/stylesheets/components/_divided.scss
neoeno: Could be — though this looks like `hr` to me — is there a reason it shouldn't be?

mconnell: Good spot, I'll get that changed

app/views/admin/feeds/show.html.erb
```diff
@@ -17,60 +17,92 @@
...
+    </div>
+
+    <ul>
+      <% @feed.sources.each do |source| %>
+        <li><%= source.name %></li>
     <% end %>
-  </ul>
-
+    </ul>
+  </div>
```
neoeno: Indentation needs a bit of attention in this file

app/views/admin/feeds/show.html.erb
```diff
@@ -17,60 +17,92 @@
...
-<h3>JS Widget URL - without headlines</h3>
-<%= text_area_tag :js_widget, script_tag_for(@feed, exclude_headlines: true) %>
-<%= link_to 'Open in a new tab', feed_url(@feed, format: :html, exclude_headlines: true), target: '_blank', class: 'button button--secondary button--small' %>
-
-<h3>JS Widget URL - with headlines</h3>
-<%= text_area_tag :js_widget, script_tag_for(@feed) %>
-<%= link_to 'Open in a new tab', feed_url(@feed, format: :html), target: '_blank', class: 'button button--secondary button--small' %>
+<div class="divided">
+  <h2>Share</h2>
+  <div class="l-4col">
```
neoeno: What's `l-4col` do?

mconnell: I think the intention was to name the parent container as there are 4 `l-4col-even` child elements within it. It works fine without the div, should I remove it? And is there a more appropriate name for each of the 4 elements that us l-4col-even?

neoeno: You're right — there probably is a better name. I'm not sure the logic behind using `even` even when there's no other possibilities, but it appears a few times and I can't think of a name so good that it overcomes the value of consistency.

Looking at the styleguide, using `.l-4col` seems consistent, but maybe add a class for it:

```
.l-4col {
  @include row;
}
```

I think that sets up the clearfix and some margins to counteract the outer edge column margins (which helps everything line up). [See here](https://github.com/moneyadviceservice/yeast/blob/master/assets/lib/_grid.scss#L44-L50)

app/views/admin/feeds/show.html.erb
```diff
@@ -17,60 +17,90 @@
...
+  </div>
+</div>
 
 <ul>
   <li><%= rule_text 'Pixel tracking', @feed.pixel_tracked? %></li>
 
   <% @feed.rules.each do |rule| %>
     <li><%= rule_text rule.label, @feed.rule_applied?(rule) %></li>
-  <% end %>
+<% end %>
```
neoeno: `  ` needed

app/views/admin/feeds/show.html.erb
```diff
@@ -17,60 +17,90 @@
...
+      <%= link_to 'Edit sources',
+        edit_admin_partner_feed_firehose_feed_sources_path(@partner, @feed), class: 'button button--secondary button--small' %>
+    </div>
+  </div>
 
   <ul>
     <% @feed.sources.each do |source| %>
       <li><%= source.name %></li>
-    <% end %>
+  <% end %>
```
neoeno: Here too

app/views/admin/feeds/show.html.erb
```diff
@@ -17,60 +17,90 @@
...
       <li class="display-list__item <%= published_class_for(item) %>">
-        <p><%= item.title %> (<%= item.published_at.to_s(:long) %>)</p>
-        <p><%= link_to_if item.published?, item.url, item.url, target: '_blank' %></p>
-      </li>
-    <% end %>
+    <p><%= item.title %> (<%= item.published_at.to_s(:long) %>)</p>
+    <p><%= link_to_if item.published?, item.url, item.url, target: '_blank' %></p>
+    </li>
+  <% end %>
   </ul>
```
neoeno: Indentation is off lines 70-76

mconnell:
Thanks @neoeno. Removed the WIP tag and updated the PR description.

asmega:
:+1:

app/views/admin/feeds/show.html.erb
```diff
@@ -17,9 +17,16 @@
...
-<h2>Rules</h2>
+<hr>
 
-<%= link_to 'manage rules', edit_admin_partner_feed_rules_path(@partner, @feed), class: 'button button--secondary button--small' %>
+<div class="heading-with-button">
+  <div class="heading-with-button__header">
+    <h2>Rules</h2>
+  </div>
+  <div class="heading-with-button__button">
+    <%= link_to 'manage rules', edit_admin_partner_feed_rules_path(@partner, @feed), class: 'button button--secondary button--small' %>
```
neoeno: 'manage rules' -> 'Manage rules' to be consistent

neoeno:
:+1:

PR #84, opened 2015-07-14
Hide unpublished content

mconnell:
Adds a method which will return all source items that are not hidden, with the exception when they are associated with a given feed.

This prevents content which is hidden from being available, but doesn't prevent content which is already associated with a feed from being visible (intended behaviour).


app/models/source_item.rb
```diff
@@ -12,6 +12,10 @@ class SourceItem < ActiveRecord::Base
   scope :by_title, -> { order(title: :asc) }
   scope :unpublished, -> { where(published: false) }
 
+  def self.visible_for_feed(feed)
+    feed.source_items | where(hidden: false)
```
asmega: does the result differ if the code is `feed.source_items.where(hidden: false)`?

mconnell: Yes. `feed.source_items.where(hidden: false)` will only return items which have a join relationship with the feed that are not hidden. The functionality requires records which are associated, regardless of whether they are hidden or not, in addition to showing unhidden records.

asmega: :+1:

asmega:
:+1:

Tr4pSt3R:
:100: :+1:

PR #85, opened 2015-07-14
firehose feeds don't have an interest in hidden source items

mconnell:
Firehose feeds shouldn't contain any hidden source items.


asmega:
:+1:

munckymagik:
:+1:

PR #86, opened 2015-07-14
change tests to not require specific array ordering

mconnell:
I've seen a handful of tests fail on me today because of the order of information being returned from the database is different to what is hardcoded in the test. This is a small change to ignore the ordering as it isn't relevant here.


spec/reports/updated_source_items_report_spec.rb
```diff
@@ -49,7 +49,8 @@
             end
 
             it 'returns the source item title, along with the two feed names' do
-              expect(subject.for_partner(partner)).to eq('Item 1' => ['Curated Feed', 'Curated Feed 2'])
+              expect(subject.for_partner(partner)['Item 1']).to include('Curated Feed')
+              expect(subject.for_partner(partner)['Item 1']).to include('Curated Feed 2')
```
neoeno: I think you could do this here:

```ruby
expect(subject.for_partner(partner)['Item 1']).to include('Curated Feed', 'Curated Feed 2')
```

munckymagik:
:+1:

Tr4pSt3R:
:100: :+1:

PR #87, opened 2015-07-14
Fetched items should be published

mconnell:
If we have hidden or unpublished content, which has been republished at source, we should unhide it and mark it as published again too.

Also did a little bit of tidy up, moving tests to a more appropriate place, and refactoring some logic out into a separate method for readability.


munckymagik:
:+1:

asmega:
:+1:

PR #88, opened 2015-07-17
add explaination why page is blank

mconnell:
Super unexciting little PR. Adds a little paragraph to reinforce that there a no unpublished items so the page doesn't look like it hasn't loaded properly.

screenshot 2015-07-17 16 17 37

screenshot 2015-07-17 16 17 30


asmega:
:+1:

urfolomeus:
:+1:

PR #89, opened 2015-07-20
Order rules

mconnell:
Allows the order which rules are applied to content within a feed to be set.

Feeds are made up of one or more source items. Feeds have rules that are applied to source items when the feed is generated. The ordering just changes the order in which those rules are applied.


app/views/admin/rule_orderings/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+<%= form_for @feed,
+             as: :feed,
+             url: admin_partner_feed_rule_ordering_path(@partner, @feed),
+             builder: Dough::Forms::Builders::Validation do |form| %>
+
+  <%= form.validation_summary %>
+
+  <%= form.fields_for :applied_feed_rules, @applied_feed_rules do |afr_form| %>
+    <%= afr_form.form_row do %>
```
neoeno: It'd be nice to provide the field name here, like `afr_form.form_row :order_id` unless there's a good reason not to. Same for the rest of this file.

app/views/admin/rule_orderings/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  <% end %>
+
+  <% if @feed.pixel_tracked? %>
+    <%= form.form_row do %>
+        <%= label_tag 'pixel_tracked_order_id', 'Pixel Tracking (fixed value)' %>
+        <%= text_field_tag 'pixel_tracked_order_id', RuleApplier::PIXEL_TRACKING_ORDER_ID, disabled: true %>
+    <% end %>
+  <% end %>
+
+  <%= form.form_row do %>
```
neoeno: Do you need a `form_row` here? I can't actually find this screen anywhere to look at when I run the server so I'm asking rather than telling!

app/views/admin/rules/edit.html.erb
```diff
@@ -9,6 +9,9 @@
 
 <h1>Manage rules</h1>
 
+<h2>Rule Ordering</h2>
+      <%= link_to 'Edit Rule Ordering', edit_admin_partner_feed_rule_ordering_path(@partner, @feed), class: 'button button--secondary button--small' %>
```
neoeno: Indentaton

urfolomeus:
@neoeno I think we've dealt with the issues you raised. As none of those form_rows were using any of the built in goodness from the Dough helper (i.e. none of them will ever have errors on those fields, the errors are at the parent object level) we've swapped them for divs that use the `form__row` class instead.

app/models/applied_feed_rule.rb
```diff
@@ -3,8 +3,20 @@ class AppliedFeedRule < ActiveRecord::Base
...
+
+  def set_default_order_id
+    return if order_id.present?
+    self.order_id = default_order_id_for_rule
+  end
+
+  def default_order_id_for_rule
+    return 20 if rule.try(:source_rule?)
+    return 40 if rule.try(:template_rule?)
+    50
```
asmega: maybe set these as methods/constants instead of magic numbersso they have some meaning

urfolomeus: @asmega we thought about that, but they are not used anywhere other than here at the moment.

neoeno:
@urfolomeus it's more of a personal approach, but I prefer to use the form helpers with the field symbols where possible. There's nothing in the method signature that suggests `form_row` is just for errors, so I'd recommend using it assuming that it does something useful or might do so in future. That's part of the power of having the style guide and the dough helpers for it

Also, having pointlessly written that, I've just checked and using `form_row` with the symbol does have a useful effect after all:

image

neoeno:
vs 
image

urfolomeus:
Hmm, I could have sworn we did that when I reviewed the code with Mark and it had no effect. Oh well. I'll add them back in, but correctly this time. Thanks.

urfolomeus:
OK @neoeno, done now.

db/schema.rb
```diff
@@ -21,6 +21,7 @@
     t.integer  "rule_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer  "order_id"
```
neoeno: The convention in rails seems (based on my intuition) to be that `something_id` is a foreign key. Unless I'm missing something, `order_id` isn't a foreign key — and I'm uncertain as to whether it qualifies as an `id` (whatever that is(. 

It's not a big deal and I know you'll be reluctant to change it given its foundational position, but I thought I'd raise it.

neoeno:
:+1:

db/schema.rb
urfolomeus: That's a good point actually. That hadn't register with me when I reviewed. Will discuss with Mark in the morning.

app/views/admin/rule_orderings/edit.html.erb
```diff
@@ -0,0 +1,42 @@
...
+
+<%= form_for @feed,
+             as: :feed,
+             url: admin_partner_feed_rule_ordering_path(@partner, @feed),
+             builder: Dough::Forms::Builders::Validation do |form| %>
+
+  <%= form.validation_summary %>
+
+  <%= form.fields_for :applied_feed_rules, @applied_feed_rules do |afr_form| %>
+    <%= afr_form.form_row(:order_id) do %>
```
benbarnett: Not seen it used like this before, cool

app/models/applied_feed_rule.rb
tomas-stefano: Even if is just used in one place, I think it's better we have a constant for this fields.

paulanthonywilson:
:+1:

tomas-stefano:
:+1:

PR #90, opened 2015-07-21
Add image author to feeds

mconnell:
Adds additional <author> and <enclosure> elements to outgoing RSS feeds for the author and image content retrieved from the MAS blog.

Also adds the ability to toggle these new elements on or off for a particular feed. By default they are applied, but there may be scenarios where this is not desired.

screenshot 2015-07-21 11 41 44


neoeno:
Is there an easy way I can add some source items so I can see all this stuff in the browser?

app/views/admin/feeds/_form.html.erb
```diff
@@ -30,6 +30,18 @@
     <%= form.label :pixel_tracked, class: 'form__label-heading--inline' %>
   <% end %>
 
+  <%= form.form_row(:show_author) do %>
+    <%= form.errors_for :show_author %>
+    <%= form.check_box :show_author %>
+    <%= form.label :show_author, 'Include <author> in RSS feed?', class: 'form__label-heading--inline' %>
```
neoeno: What's this class do?

neoeno: @spencer I have been referred to you on this question!

spencer: moves the label inline next to the checkbox, rather than above it I believe.

neoeno: Do you know where it's defined? I can't seem to find it anywhere — removing it doesn't seem to have any obvious effect either

spencer: I guess I found it here:
https://github.com/moneyadviceservice/frontend-sandbox/blob/3306a07365fdaa7239ce22d4dc6d6e24898a2a00/source/stylesheets/components/common/_form.scss

But maybe in the many branch/merges the actual code got lost? Or maybe I never imported it.

urfolomeus: @spencer @neoeno are there any actions to come out of this conversation?

urfolomeus:
@neoeno you can run `rake webfeeds:fetch:all` to get content to play about with for this. It takes a minute or two to run though.

app/views/admin/feeds/_form.html.erb
benbarnett: loading-pup

urfolomeus: Must. Look. Away.

neoeno: It's your PR — you decide! Seems like the class doesn't do anything, so either it should be removed or whatever it was for should be added to the CSS. Suggesting the former, since presumably its not being in the CSS hasn't caused any problems?

urfolomeus: Sounds fair. You happy enough for me to remove that @spencer? We can always add it back later if it turns out to have been needed?

urfolomeus: OK @neoeno @benbarnett @spencer took an executive decision and removed it. Can haz :+1: now? :)

db/schema.rb
```diff
@@ -21,6 +21,7 @@
     t.integer  "rule_id"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer  "order_id"
```
neoeno: Should this be here?

neoeno: Looks like some of the schema changes from this PR leak into the other one too.

urfolomeus: Hmm yeah. I've had this happen before. Thanks for raising. Will check tomorrow morning.

neoeno:
:+1: looks good

benbarnett:
:+1:

PR #91, opened 2015-07-22
Filter by source

urfolomeus:
When selected items to add to a curated feed, the items can be filtered based on their originating source.


app/views/admin/feed_source_items/edit.html.erb
```diff
@@ -10,20 +10,23 @@
...
+  <%= form_for @feed,
+               as: :feed,
+               url: admin_partner_feed_feed_source_items_path(@partner, @feed),
+               builder: Dough::Forms::Builders::Validation do |form| %>
+
+    <%= render 'source_item_checkboxes', form: form, source_items: @source_items %>
+
+    <div class="form__action-row">
+      <%= link_to 'cancel', admin_partner_feed_path(@partner, @feed), class: 'form__action-row-cancel' %>
+      <%= submit_tag 'Save', class: 'button button--primary' %>
```
benbarnett: Is there a difference between `submit_tag` and `form.submit`? If not, we should probably pick one for consistency. (ref: https://github.com/moneyadviceservice/webfeeds/pull/91/files#diff-944359b05b20db35ba1156ad597c732cR11)

urfolomeus: Ah, good catch! This was originally a form_tag_helper and not a form_helper. Missed the submit though.

benbarnett:
:+1:

app/controllers/admin/feed_source_items_controller.rb
```diff
@@ -5,7 +5,7 @@ def new
   end
 
   def create
-    if @feed.select_source_items(allowed_params[:source_item_ids])
+    if @feed.select_source_items(source_item_data)
```
tomas-stefano: Is not better to be a method with a question mark?

app/controllers/admin/feed_source_items_controller.rb
```diff
@@ -16,7 +16,7 @@ def edit
   end
 
   def update
-    if @feed.select_source_items(allowed_params[:source_item_ids])
+    if @feed.select_source_items(source_item_data)
```
tomas-stefano: Same here.

app/controllers/admin/feed_source_items_controller.rb
```diff
@@ -25,14 +25,18 @@ def update
...
 
-  def allowed_params
-    params[:feed][:source_item_ids].pop
-    params.require(:feed).permit(source_item_ids: [])
+  def source_item_data
+    params.require(:feed).require(:source_item_data).permit!
+  end
+
+  def filter_options
+    return {} unless params.key?(:filter)
```
tomas-stefano: Just a sanity check. Did we have a spec for this case?

app/controllers/admin/feed_source_items_controller.rb
urfolomeus: This method is not a boolean method, even though it does return true or false. It adds source_items to a feed. We work to the convention that ? methods do not have side effects.

app/helpers/filter_helper.rb
```diff
@@ -0,0 +1,6 @@
+module FilterHelper
+  def source_selected?(obj_id)
+    return false unless params[:filter].try(:[], :source_id)
+    params[:filter][:source_id].include?(obj_id)
```
tomas-stefano: Maybe I misunderstood, but what do you think of:
 
`params[:filter].try(:[], :source_id).present? && obj_id.in?(params[:filter][:source_id])` ?

You can move the params[:filter].try(:[], :source_id) to a variable if you like.

app/models/curated_feed.rb
```diff
@@ -3,17 +3,24 @@ class CuratedFeed < Feed
...
-        feed_source_items.destroy_all
+  def select_source_items(source_item_data)
+    add_selected_source_items(source_item_data)
+    remove_deselected_source_items(source_item_data)
+  end
+
+  private
+
+  def add_selected_source_items(source_item_data)
+    new_ids = ids_with_state(source_item_data, '1') - source_item_ids.map(&:to_s)
```
tomas-stefano: Why '1'? I know that is because the params send as '1' when is selected, but can we move this number to a variable?

app/models/curated_feed.rb
```diff
@@ -3,17 +3,24 @@ class CuratedFeed < Feed
...
+    source_items << SourceItem.where(id: new_ids)
+  end
+
+  def remove_deselected_source_items(source_item_data)
+    ids_to_remove = ids_with_state(source_item_data, '0')
```
tomas-stefano: The same for the '0'.

app/views/admin/feed_source_items/_source_item_checkboxes.html.erb
```diff
@@ -0,0 +1,10 @@
+<% source_items.each do |source_item| %>
+  <div class="form__row">
+    <%= form.check_box :_destroy, { name: "feed[source_item_data][#{source_item.id}]", checked: @feed.source_item_ids.include?(source_item.id) }, '1', '0' %>
```
tomas-stefano: Are we using presenters? Can we move the checked part and the parameter name to it What do you think?

app/controllers/admin/feed_source_items_controller.rb
tomas-stefano: Ok then. :)

app/controllers/admin/feed_source_items_controller.rb
mconnell: Yep. It's a private method so it's not tested directly, but any spec which doesn't explicitly set the filter key covers the behaviour https://github.com/moneyadviceservice/webfeeds/blob/filter_by_source/spec/controllers/admin/feed_source_items_controller_spec.rb#L15

tomas-stefano: Ok then, :)

app/views/admin/feed_source_items/_source_item_checkboxes.html.erb
mconnell: We aren't using a presenter pattern in the admin section currently.

tomas-stefano: Ok.If I change the name or the checked, probably something will not work. Is there are tests for it? Or is this consider important to have a test?

mconnell:
@tomas-stefano I think we've resolved all the requested changes now. We decided to use a helper method instead of a presenter for the checkbox options.

tomas-stefano:
Good work mate. :+1: :+1: (squash the commits and everything good to go)

PR #92, opened 2015-07-24
Dynamically select/deselect source items on a feed

mconnell:
This PR changes the way source items are selected for feeds.

Previously, you would check a number of checkboxes and then press a save button at the bottom of the page. Now, checking a checkbox will automatically add or remove the source item on the feed.


app/assets/javascripts/application.js
```diff
@@ -13,5 +13,5 @@ require(['jquery', 'componentLoader'], function($, componentLoader) {
 require([], function() {
   'use strict';
 
-  require(['PixelTrackerGenerator', 'RuleForm']);
+  require(['PixelTrackerGenerator', 'RuleForm', 'SourceItemToggler']);
```
neoeno: Don't think this is required — Dough's `componentLoader` will load it for you on demand.

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+  function SourceItemToggler($el, config) {
+    SourceItemToggler.baseConstructor.call(this, $el, config, defaultConfig);
+  }
+
+  SourceItemToggler.componentName = 'SourceItemToggler';
+
+  /**
+   * Inherit from base module, for shared methods and interface
+   * @type {[type]}
+   * @private
```
neoeno: You can remove these last two JSDoc tags, as it's not a function so they're inapplicable

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+   */
+  DoughBaseComponent.extend(SourceItemToggler);
+  SourceItemTogglerProto = SourceItemToggler.prototype;
```
neoeno: I think the new standard is to add a `componentName` too — see https://github.com/moneyadviceservice/rad/blob/master/app/assets/javascripts/modules/ConfirmableForm.js#L37-L39

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+
+  /**
+   * Init function
+   *
+   * Set up listeners
+   * @param  {boolean} initialised Successfully initialised?
```
neoeno: `initialised` here isn't a boolean — see https://github.com/moneyadviceservice/rad/blob/master/app/assets/javascripts/modules/ConfirmableForm.js#L41-L54

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+   * @return {object}              SourceItemTogglerProto
```
neoeno: Return value here is actually `SourceItemToggler`

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+   */
+  SourceItemTogglerProto.init = function(initialised) {
+    var $checkboxes = $(this.$el.find('input[type="checkbox"]'));
```
neoeno: You shouldn't need to wrap with `$()` here, since `find` will return a wrapped collection istelf.

app/assets/javascripts/modules/SourceItemToggler.js
urfolomeus: Line 28 does this :)

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+    return this;
+  };
+
+  /**
+   * Handle toggle event
+   *
+   * @param {object} e triggered event data
+   * @private
+   */
+  SourceItemTogglerProto._doToggle = function(e) {
```
neoeno: This method could possibly be named better. Maybe `_handleCheckboxToggle` or `_handleSourceItemToggle`? I'm mostly preferring `handle` to `do` here

app/assets/javascripts/modules/SourceItemToggler.js
```diff
@@ -0,0 +1,96 @@
...
+   * Init function
+   *
+   * Set up listeners
+   * @param  {boolean} initialised Successfully initialised?
+   * @return {object}              SourceItemTogglerProto
+   */
+  SourceItemTogglerProto.init = function(initialised) {
+    var $checkboxes = $(this.$el.find('input[type="checkbox"]'));
+
+    this.url = this.$el.data('dough-source-item-toggler-url');
```
neoeno: I think MAS prefer to have these attribute names extracted out to the defaultConfig — e.g. https://github.com/moneyadviceservice/rad/blob/master/app/assets/javascripts/modules/MultiTableFilter.js#L20-L25 (worth checking examples of how it's used throughout that file too)

neoeno:
I'm not sure if the user feedback is adequate here. It assumes that the server gets the request before the user navigates away from the page (cancelling the request). Worst case is the user selects an item and then navigates away from the page, thinking they've selected the item when they haven't.

Further:

* A user who isn't aware this is an AJAX form (nothing in the interface suggests this) might be surprised to find these changes actually being applied, or might be confused by not having to submit the form.
* A user who is a bit technical and knows there is a delay in the process, will know they need to check the data has been saved successfully because they haven't received confirmation that it has done so.

A simple way of signalling 'in progress' would be as follows:

1. The user selects the checkbox
2. Your script catches the event, marks the checkbox disabled (visual change to signal in-progress), and submits the AJAX request.
3. Your script catches the server response, un-disables the checkbox, and reports failure if needed.

There are nicer ways of doing it, but that's a relatively intuitive MVP version.

mconnell:
@neoeno Added the disabling/enabling of checkboxes behaviour

neoeno:
:+1: looks good

mconnell:
Just rebased this branch as it has started to fall behind master.

asmega:
the ruby stuff looks fine :+1:

app/views/admin/feed_source_items/_source_item_checkboxes.html.erb
```diff
@@ -1,15 +1,17 @@
...
+<div data-dough-component="SourceItemToggler" data-dough-source-item-toggler-url="<%= admin_partner_feed_feed_source_items_path(feed.partner, feed) %>">
+  <% source_items.each do |source_item| %>
+    <div class="form__row">
+      <%= form.check_box "source_item_#{source_item.id}", check_box_options(feed, source_item), '1', '0' %>
+      <%= form.label "source_item_#{source_item.id}", class: published_class_for(source_item) do %>
+        <%= source_item.title %> (<%= source_item.published_at.to_s(:long) %>)
+        <br>
+        <%= link_to_if source_item.published?, source_item.url, source_item.url, target: '_blank' %>
       <% end %>
+      <div>
```
benbarnett: Is this div needed for anything?

benbarnett:
Looks good :+1:

urfolomeus:
Thanks @neoeno @asmega and @benbarnett :)

PR #93, opened 2015-07-27
Don't add an AppliedFeedRule to the Feed for the PixelTracker

mconnell:
This PR changes a collection form being an ActiveRecord association collection, to simply an array. The is an AppliedFeedRule that should be pushed into the collection, but the side effect of `<<` on an AR collection is it attempts to persist the object which we don't want, coercing to an `Array` mitigates this.


rnalexander:
:+1:

darokel:
:+1:

mconnell:
Thanks @rnalexander @darokel

PR #96, opened 2015-07-28
Adding a special rule for prepending canonical links to feed items

mconnell:
Adds an option to feeds which enables a canonical link HTML tag to be included within each item of content in a feed. It is enabled by default.


asmega:
:+1:  What happens when the order value is the same (e.g. 99)?

urfolomeus:
If rules have the same ordering they will be run in the order in which the database returns them. What we are saying by giving two or more rules the same ordering is that those rules have no precedence over one another.

asmega:
@urfolomeus Thanks, good to know

benbarnett:
:+1:

PR #97, opened 2015-07-28
Add feed fixen

urfolomeus:
Through a previous feature we had accidentally created a rare situation where, if you had no source rules, it was possible to get stuck on the apply rules page. This PR fixes that issue.


app/models/feed.rb
```diff
@@ -16,11 +16,12 @@ def applicable_source_rules
   end
 
   def update_applied_rules(selected_rule_ids)
+    return false if selected_rule_ids.nil?
+
     remove_rules missing_rule_ids(selected_rule_ids)
```
Tr4pSt3R: there's no test to clarify the definition of a 'missing rule' and to test that they can be successfully removed if they are indeed missing. 
Apart from this all else looks spotless

urfolomeus: Thanks @Tr4pSt3R. The intention of the context starting on line 199 is to test this. Is it not enough?

neoeno:
:+1:

app/models/feed.rb
Tr4pSt3R: ```context 'feed has applied rule, not in array' do``` yup that looks fine. What is array? I know it's not part of your changes but is there a better name for it please?
Everything else is fine
:+1:

urfolomeus: Sounds good. Will fix just now, thanks.

urfolomeus: Done @Tr4pSt3R.

Tr4pSt3R:
@urfolomeus :tada: it's a ton more easier to understand. Thanks :+1: :ship:

urfolomeus:
Thanks for all your feedback, @Tr4pSt3R. Appreciated :)

PR #98, opened 2015-07-28
Adding some tests to ensure that capture behaviour remains in the app…

urfolomeus:
… if the transform implementation is changed


asmega:
it seems odd that the content contains the less/greater than signs, why doesn't the regex handle that instead?

urfolomeus:
@asmega the target content that we're using here is `<p>first</p><div class="callout">second</div><p>third</p>` (not obvious from the diff, granted). What I'm doing is getting the "first", "second" and "third" text by finding all `\w` chars between the angle braces and capturing it. However, when I come to do the replace, I need to add the angle braces back in as they're not in the capture, and nor can they be as I need the replacement text to be between them.

I'm not sure I'm making much sense :)

Maybe I should have used a more simple example, but this is pretty close to the end use case.

neoeno:
You could use lookarounds to check that the angle brackets are there without including it in the content. But that's not relevant to the efficacy of these tests I guess. Plus I can't actually get it to work nicely in 10 minutes so it must be impossible

:+1:

asmega:
- this is maybe what you are after?
- I used `+` instead of `*` as I'm guessing you don't want to capture empty tags? If not, the `*` can go back. 
- regexes don't have to be a string if they are between the slashes it is simply a regex object.

```
diff --git a/spec/models/rule_spec.rb b/spec/models/rule_spec.rb
index d716112..04f025c 100644
--- a/spec/models/rule_spec.rb
+++ b/spec/models/rule_spec.rb
@@ -225,13 +225,13 @@ RSpec.describe Rule, type: :model do

     context 'captures' do
       it 'includes the captured material if reused in the replace' do
-        rule   = Rule.new(rule_type: 'transform', content: '>\1 word</', regex: '>(\w*)</')
+        rule   = Rule.new(rule_type: 'transform', content: '\1 word', regex: /(?<=>)(\w+)(?=<)/)
         result = rule.apply_to(target)
         expect(result).to eq('<p>first word</p><div class="callout">second word</div><p>third word</p>')
       end

       it 'does not include the captured material if not reused in the replace' do
-        rule   = Rule.new(rule_type: 'transform', content: '> word</', regex: '>(\w*)</')
+        rule   = Rule.new(rule_type: 'transform', content: ' word', regex: /(?<=>)(\w+)(?=<)/)
         result = rule.apply_to(target)
         expect(result).to eq('<p> word</p><div class="callout"> word</div><p> word</p>')
       end
@@ -239,13 +239,13 @@ RSpec.describe Rule, type: :model do

     context 'named captures' do
       it 'includes the captured material if reused in the replace' do
-        rule   = Rule.new(rule_type: 'transform', content: '>\k<count> word</', regex: '>(?<count>\w*)</')
+        rule   = Rule.new(rule_type: 'transform', content: '\k<count> word', regex: /(?<=>)(?<count>\w+)(?=<)/)
         result = rule.apply_to(target)
         expect(result).to eq('<p>first word</p><div class="callout">second word</div><p>third word</p>')
       end

       it 'does not include the captured material if not reused in the replace' do
-        rule   = Rule.new(rule_type: 'transform', content: '> word</', regex: '>(?<count>\w*)</')
+        rule   = Rule.new(rule_type: 'transform', content: ' word', regex: /(?<=>)(?<count>\w+)(?=<)/)
         result = rule.apply_to(target)
         expect(result).to eq('<p> word</p><div class="callout"> word</div><p> word</p>')
       end
```

urfolomeus:
Sorry @asmega but I don't think this is any more readable. `'>(\w*)</'` is the way that Edmund has been writing these, so we followed his style. `/(?<=>)(\w+)(?=<)/` might be more "correct", but both solutions will actually test the implementation in the same way, no? :)

urfolomeus:
It's also worth pointing out that this doesn't change any behaviour, nor test anything new.

asmega:
sure. :+1:

urfolomeus:
Thanks.

FWIW I learned more about regex captures through this, so thanks. :)

PR #102, opened 2015-07-29
add unpublished_url text field

mconnell:
Adds a text field to allow the unpublished url for a source to be edited

Before:
screen shot 2015-07-29 at 10 23 27

After:
screen shot 2015-07-29 at 10 23 11


neoeno:
:+1: what's an unpublished URL?

mconnell:
A Source has a URL containing all of the content that is publicly available (published). For some sources, there is an additional URL that lists all of the content which has been removed from being available (unpublished).

In the case where an unpublished url exists on a source, there is a nightly task which will mark content persisted in the application as unpublished.

Tr4pSt3R:
:+1:

PR #103, opened 2015-07-29
Delete system rules

urfolomeus:
This functionality allows us to delete system rules from Webfeeds. A system rule is one that is available to all feeds and not just an individual feed. Source rules work on all of a feed's items that come from an affected source, whereas template rules work on all of a feed's items regardless of their source.

In order to do this we lifted the ConfirmableForm component from the RAD project. This is in the process of being PRd into Dough, but we will refactor to that solution once it is available in a separate PR.


app/controllers/admin/source_rules_controller.rb
```diff
@@ -28,6 +28,12 @@ def update
     end
   end
 
+  def destroy
+    rule = Rule.source_rules.find(params[:id])
+    rule.destroy
```
asmega: if you are not checking the result ie success/failure i'd call `.destroy!` instead

urfolomeus: Fair point, @asmega. Will update now.

urfolomeus:
@asmega done

app/controllers/admin/source_rules_controller.rb
urfolomeus: @asmega done.

asmega:
:+1:

urfolomeus:
We've now swapped out our local version of ConfirmableForm for the Dough one. This does not affect any other files.

Tr4pSt3R:
Travis is failing. This seems to be a common failure. Probably an unmatched bracket somewhere


```ruby
1) widget previewing when headlines are excluded when the feed has one source item
     Failure/Error: Unable to find matching line from backtrace
     ActionView::Template::Error:
       couldn't find file 'dough/assets/js/lib/components/ConfirmableForm'
         (in /home/travis/build/moneyadviceservice/webfeeds/app/assets/javascripts/require_config.js.erb:7)
     # ./app/views/layouts/application.html.erb:51:in `_app_views_layouts_application_html_erb___4198187465288910535_72048520'
     # ./app/controllers/feeds_controller.rb:8:in `show'
```

urfolomeus:
Thanks, @Tr4pSt3R. Caught and fixed. Was a typo on my part. :)

@asmega this shouldn't have changed anything that you reviewed other than to swap out our local implementation of ConfirmableForm, but you may want to give it a quick rescan?

asmega:
:+1:

Tr4pSt3R:
:100: :+1:

app/assets/stylesheets/components/_button.scss
```diff
@@ -3,3 +3,23 @@
   padding: 14px 16px;
   line-height: 0.5;
 }
+
+.button--red {
```
neoeno: I think the existing dough buttons are named semantically. So maybe `button--danger` would be better? @benbarnett?

benbarnett: Yep concur

app/assets/stylesheets/components/_system_rule.scss
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+
+.system-rule__actions {
+  @include column(12);
+  @include respond-to($mq-m) {
+    @include column(4);
+  }
+
+  text-align: right;
```
neoeno: I'd organise it this way:

```sass
.system-rule__actions {
  @include column(12);
  text-align: right;
  
  @include respond-to($mq-m) {
    @include column(4);
  }

  .button {
    float: right;
    margin-top: $baseline-unit;
    margin-left: $baseline-unit;
  }
}
```

That keeps the styles at each breakpoint together.

mconnell:
@neoeno @benbarnett Thanks, I've amended the code with those suggestions :smiley:

neoeno:
:+1:

mconnell:
Thanks for the everyone!

PR #104, opened 2015-07-30
Pull in design work from stale branch

mconnell:
We had some design changes sitting on an old branch that never made it to PR earlier. The branch was quite far behind, so the commits have effectively been rounded up, cherry picked, reviewed and squashed. The PR is a bit chunky, but it's mostly due to shuffling around the structure and adding in new styles.

In Summary:

* The `nested_form` gem was removed and contact creation was moved into its own controller.
* Content on the page has been re-jigged. Feeds would appear at the bottom of the page, yet were the most important detail. Partner information and contacts are moved into the side bar as secondary information. Each portion/component has also been extracted out into a partial as well to clean up the `show.html.erb` template.
* The pixel tracker component was originally in the sidebar too, but if a partner does not have a lot of feeds, it's likely that the pixel tracking functionality is going to be used frequently, so there was a preference to have it directly under the list of feeds.

Before:
screen shot 2015-07-30 at 14 12 50
(feeds are under the 'fold' on a 27" monitor)
screen shot 2015-07-30 at 14 13 06

After:
screen shot 2015-07-30 at 13 48 18


pmcostello:
:+1: looks good

benbarnett:
Nice :+1:

PR #108, opened 2015-08-14
Adding simple validation to the regex to help Edmund catch errors.

urfolomeus:
Before this fix, if Edmund adds a rule with an invalid regular expression matcher, the application will 500 with no obvious error. This fix will instead validate the regex on save to ensure that any errors are caught before the can affect the feeds.


mconnell:
:+1:

darokel:
:+1:

urfolomeus:
Thanks @mconnell and @darokel :)

PR #110, opened 2015-11-13
Adding default author to RSS feeds

johnplayer1982:
Adding a default author of 'The Money Advice Service' when a user selects 'Include authors' when creating a new feed, but an author doesn't exist


darokel:
How you wrote it the first time was fine. You don't need the second commit.

johnplayer1982:
Thanks for taking a look, The first commit adds the default author even if the 'include authors' option isn't selected.  The default author needs to be applied when both the include authors option is selected and there is no author stored for the feed item

darokel:
Since the `@feed.show_author?` is being relied on for both conditions, you could do something like this:

```ruby
if @feed.show_author?
  return @source_item.author.present?
  default_author
end
```

`return @source_item.author.present?` will return true and exit the method if it's present and not blank, otherwise this line will be ignored and the `default_author` will be set.

app/presenters/source_item_presenter.rb
```diff
@@ -64,7 +64,15 @@ def include_image?
   end
 
   def include_author?
-    @feed.show_author? && @source_item.author.present?
+    if @feed.show_author? && @source_item.author.present?
+      true
+    elsif @feed.show_author? && @source_item.author.blank?
+      default_author
+    end
```
tomas-stefano: Why the method has a question mark if is returning default_author on the second condition?

app/presenters/source_item_presenter.rb
```diff
@@ -64,7 +64,15 @@ def include_image?
...
+  end
+
+  def default_author
+    @source_item.author = 'The Money Advice Service'
```
tomas-stefano: I would put the conditions here for this assignment and leave the include_author? intact. (For at least I understand the feature).

app/presenters/source_item_presenter.rb
johnplayer1982: Hi Tomas, It is because the user has the option to include an author in an RSS feed, if the option isn't selected neither the author or default author is output to the RSS, the default author will only be used if the user chooses to include an author and none exists.

tomas-stefano: ok. Can you change to:

```ruby
return false unless @feed.show_author?

@source_item.author.present? || default_author
````

Comment (johnplayer1982) [inline app/presenters/source_item_presenter.rb]: Works beautifully and much less code, thanks @tomas-stefano

File: app/presenters/source_item_presenter.rb [hunk 3]
```diff
@@ -64,7 +64,12 @@ def include_image?
   end
 
   def include_author?
-    @feed.show_author? && @source_item.author.present?
+    return false unless @feed.show_author?
+    @source_item.author.present? || default_author
+  end
+
+  def default_author
+    @source_item.author = 'The Money Advice Service'
```
paulanthonywilson: There's something not right about modifying `@source_item` during a query method. (https://en.wikipedia.org/wiki/Command–query_separation).

I'd suggest adding an author method, something like

```
def include_author? # or maybe just inline this method
   @feed.show_author?
end

def author
  @source_item.author || 'The Money Advice Service'
end
```

Then at line 49

```
   xml.author  author if include_author?
```

johnplayer1982: Thanks for the advice @paulanthonywilson

paulanthonywilson:
I'd also think about adding something to [the spec](https://github.com/moneyadviceservice/webfeeds/blob/add-default-author/spec/presenters/source_item_presenter_spec.rb) to deal with the case of using a default author.

paulanthonywilson:
Sorry about the timing. Something like replacing the context at line 275 in the spec. (YMMV some people might not like the hard coded text of the default author there.)

```
    context 'no author, but feed is set to include an author' do
      before :each do
        source_item.author = nil
        feed.show_author = true
      end

      it 'excludes <author> from the xml' do
        expect(subject.to_xml).to include('<author>The Money')
      end
    end
```

johnplayer1982:
Hi Paul, thanks for the tip on the spec! Spec now added

paulanthonywilson:
Great :-)

:+1:

mconnell:
:+1:

PR #114, opened 2016-04-20
Adding rollbar

urfolomeus:
This PR adds Rollbar exception reporting to the rake tasks that pull content from the various MAS sources. We had an issue where the blog feed from publify was 500ing and we didn't notice until a partner pointed it out to us. In future errors like this will be sent to the partnership team via Rollbar.


munckymagik:
👍

urfolomeus:
Closing so that I can adopt the new PR approach

PR #117, opened 2016-05-04
Add a rake task changing the blog source data

tomas-stefano:
## Context

Blog source urls are going to change with the blog domain name change. MAS needs to ensure that the RSS Article web feed that is used by big partners like MSN, AOL etc. works. The RSS feed is used by partners to pick/select the content they want to put on their site.

## Wait

**Do not run this task before the blog is migrated.**


urfolomeus:
@tomas-stefano I'd run this on staging first to make sure that you don't get duplicate entries for the blog with the URL changing. I don't think you should as the blog sets its on guid. But still ... :)

tomas-stefano:
I don't have access to staging/production yet.

jiannis:
OK - looks good - this is part of the [final migration steps of the blog (card #7315)](https://moneyadviceservice.tpondemand.com/entity/7315) to be executed then.

tomas-stefano:
I changed to dev approved to do not show on the channels. I added this PR to the user story. I will revisit when the other blog user stories is done.

tomas-stefano:
I changed to WIP because since there are more acceptance criteria on the user story than before, I will make some tests with webfeeds staging.

tomas-stefano:
**TL; DR I am waiting for re-thumbs**

Investigating the webfeeds source code only changing the url to point to the new blog domain, the webfeed will fetch the blog rss and updating all the links. So we need to make sure that the blog rss is generating the right links.

About the webfeeds source code you can see more details here:
https://github.com/moneyadviceservice/webfeeds/blob/master/app/models/source.rb#L16
https://github.com/moneyadviceservice/webfeeds/blob/master/app/models/source.rb#L41
https://github.com/moneyadviceservice/webfeeds/blob/master/app/models/source_item.rb#L23
https://github.com/moneyadviceservice/webfeeds/blob/master/app/models/source_item.rb#L49

Tr4pSt3R:
👍

